PR #3, opened 2019-12-04
Add Dockerfile for CI usage

JAORMX:
Similarly to the one we added for the compliance-operator, this adds a
dockerfile for CI use only.


Dockerfile.ci
```diff
@@ -0,0 +1,31 @@
+# Step one: build file-integrity-operator
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
```
mrogers950: golang-1.13

JAORMX: there is no such tag yet in CI. This failed for the compliance-operator.

PR #7, opened 2019-12-04
Add FileIntegrityStatus controller

mrogers950:
This adds a Phase field to FileIntegrityStatus and a controller to manage it. The phases are

* Initializing - The file integrity checking is initializing: The controller sets this when the re-initialization daemonSet exists (upon config change), or the AIDE runner daemonSet is not totally ready.
* Active - The file integrity checking is currently active: The controller sets this when the AIDE runner daemonSet has available pods on all nodes.
* Inactive - There is no active integrity checking: The controller sets this when the AIDE runner daemonSet does not exist.

The controller reconciles every 30 seconds. The reason there is a continuous reconcile loop is because there is no informer action triggered by a daemonSet status update, which is what the controller monitors for status changes. So there is an initial start with the CR creation, and then it loops.

### DIFF
--- a/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
+++ b/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
@@ -7,6 +7,14 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type FileIntegrityStatusPhase string
+
+const (
+	PhaseInitializing FileIntegrityStatusPhase = "Initializing"
+	PhaseActive       FileIntegrityStatusPhase = "Active"
+	PhasePending      FileIntegrityStatusPhase = "Pending"
+)
+
 // FileIntegritySpec defines the desired state of FileIntegrity
 // +k8s:openapi-gen=true
 type FileIntegritySpec struct {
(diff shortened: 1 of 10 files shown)


pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -7,6 +7,14 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type FileIntegrityStatusPhase string
+
+const (
+	PhaseInitializing FileIntegrityStatusPhase = "Initializing"
+	PhaseActive       FileIntegrityStatusPhase = "Active"
+	PhaseInactive     FileIntegrityStatusPhase = "Inactive"
```
JAORMX: Lets change this one to Pending instead of Inactive.

tests/e2e/e2e_test.go
```diff
@@ -1,16 +1,68 @@
...
+!/hostroot/etc/.*~
+!/hostroot/etc/kubernetes/static-pod-resources
+!/hostroot/etc/kubernetes/aide.*
+!/hostroot/etc/docker/certs.d
+!/hostroot/etc/selinux/targeted
+
+# Catch everything else in /etc
+/hostroot/etc/    CONTENT_EX`
+
 func TestFileIntegritySingleRun(t *testing.T) {
```
JAORMX: Let's rename the test now that it's more extensive. And add a doc that describes it.

PR #9, opened 2019-12-05
default GOPROXY_SETTINGS to empty

mrogers950:
Trying this out for CI (https://github.com/openshift/release/pull/6242)


JAORMX:
Have you tried it out? We need those at least for the go build commands

JAORMX:
Look at the commit history and revert my patch that introduced it

JAORMX:
I think we can abandon this one.

PR #17, opened 2019-12-11
add an e2e test for a missing AIDE configmap

mrogers950:
This test just ensures there's no re-init when you specify a nonexistant AIDE configmap, plus some refactoring.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	goctx "context"
 	"testing"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -13,6 +14,7 @@ import (
 
 	fileintv1alpha1 "github.com/openshift/file-integrity-operator/pkg/apis/fileintegrity/v1alpha1"
 	"github.com/openshift/file-integrity-operator/pkg/common"
+	"github.com/openshift/file-integrity-operator/pkg/controller/fileintegrity"
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
 )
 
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_test.go
```diff
@@ -210,3 +188,80 @@ func waitForScanStatus(t *testing.T, f *framework.Framework, namespace, name str
...
+	f, testctx, namespace := setupTest(t)
+	defer testctx.Cleanup()
+
+	createTestConfigMap(t, f, testIntegrityName, testConfName, namespace, testConfDataKey, testAideConfig)
+
+	// wait for an initialization period.
+	err := waitForScanStatus(t, f, namespace, testIntegrityName, fileintv1alpha1.PhaseInitializing)
+	if err != nil {
+		t.Error(err)
+	}
```
JAORMX: should we remove this block and just straight up wait for the Active phase?

tests/e2e/e2e_test.go
```diff
@@ -210,3 +188,80 @@ func waitForScanStatus(t *testing.T, f *framework.Framework, namespace, name str
...
+	defer testctx.Cleanup()
+
+	// Non-existent conf
+	updateFileIntegrityConfig(t, f, testIntegrityName, "fooconf", namespace, "fookey")
+
+	// No re-init should happen, let this error pass.
+	err := waitForScanStatusWithTimeout(t, f, namespace, testIntegrityName, fileintv1alpha1.PhaseInitializing, time.Second*5, time.Second*30)
+	if err == nil {
+		t.Errorf("status changed to initialization in error")
+	}
```
JAORMX: same comment as above.

tests/e2e/e2e_test.go
mrogers950: Not here, because we want to be sure a configmap change flips status to Initializing (this usually takes longer than the first initialization because of the rolling update)

tests/e2e/e2e_test.go
mrogers950: Ditto

mrogers950: For this one, we take a short time to ensure that the status does _not_ change to initializing, since the missing configmap update is ignored.

tests/e2e/e2e_test.go
```diff
@@ -107,82 +103,64 @@ func TestFileIntegrityConfigurationStatus(t *testing.T) {
...
 
 	// wait for an initialization period.
-	err = waitForScanStatus(t, f, namespace, testIntegrityName, fileintv1alpha1.PhaseInitializing)
-	if err != nil {
-		t.Error(err)
-	}
+	//err = waitForScanStatus(t, f, namespace, testIntegrityName, fileintv1alpha1.PhaseInitializing)
+	//if err != nil {
+	//	t.Error(err)
+	//}
```
JAORMX: maybe we should just remove this block

PR #18, opened 2019-12-12
Make `all` target only build

JAORMX:
We already trigger the unit tests and verifications in their own targets
in CI. So no need to call them when building the binary.


JAORMX:
/retest
error setting up cluster (not related)

PR #14, opened 2019-12-10
bump(*): Vendor dependencies

tnozicka:
This ensures offline builds and ensures the project requirements are met.

Also the project level golang builder is about to be switched to force vendoring mode.

/priority critical-urgent
/assign jhrozek JAORMX mrogers950

PR #19, opened 2019-12-12
vendor dependencies

JAORMX:
This adds all the changes required to vendor our dependencies. It's pased on @tnozicka's https://github.com/openshift/file-integrity-operator/pull/14 PR which introduces this in the first place, but this adds the missing gosec dependency and changes to the makefile to make this work.


JAORMX:
etcd call timed out...

JAORMX:
rebased on top of the PR that introduced an extended timeout

go.mod
```diff
@@ -4,16 +4,9 @@ go 1.13
 
 require (
 	github.com/go-openapi/spec v0.19.0
-	github.com/onsi/ginkgo v1.10.3 // indirect
-	github.com/onsi/gomega v1.7.1 // indirect
```
mrogers950: Why are these removed? I think that if you run `go mod tidy` and then don't run `make e2e` you miss out on these deps. If you run `make e2e` at least once and then vendor, these will get included.

JAORMX: I don't find where the usage of ginkgo is.

PR #20, opened 2019-12-12
Up timeout to 15 min

JAORMX:
The initialization stage is taking too long for some reason. Lets pu the
timeout to 15 min to reduce the errors in CI while we fix this.


jhrozek:
Trivial patch, can be merged once tests pass
/lgtm

PR #21, opened 2019-12-12
Delete DS pods instead of triggering a rolling update

mrogers950:
Triggering a rolling update to re-initialize the AIDE pods can take some time.
Delete the DS pods instead. 

Trying this out with CI


mrogers950:
@JAORMX I fixed this one up, PTAL.

JAORMX:
@mrogers950 great! this should be faster. Have you tested it locally too?

PR #22, opened 2019-12-12
e2e: poll for configMap changes instead of status

mrogers950:
Rather than polling to check if we transitioned to Initializing and back during config updates, we just poll for Active status and the configMap result that we expect.


JAORMX:
/lgtm

I like this better for the end to end tests. We shouldn't care about the inner workings of the operator. Just the reuslt. For the inner bits we should have unit tests instead.

PR #15, opened 2019-12-10
WIP: Add status condition types

mrogers950:
Just sketching out the types for now.

PR #25, opened 2019-12-16
Add status condition types

JAORMX:
This builds on @mrogers950 work in #15 to add the status conditions.

Note that this only adds the types and refactors a little of the logic to make things easier.

A subsequent PR will introduce the logic and the logcollector.


jhrozek:
These patches look quite good to me. I have some very minor questions:
 1. Why is using `github.com/go-logr/logr` directly preferred over using the wrapper over it from `https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/log.go`? This is not so much related to this patchset, but more of FMI.
 2. Seeing that there used to be the struct `FileIntegrityStatusConditions` made me think if it was meant for extension or shared status between structs that wrap or contain it. Otherwise the substitution seems sane to me.
3. Do we want to merge patches with WIP in the subject? :)

Since the tests passed, lgtm can be added, just the WIP in git history would trigger my OCD and also I guess it's easier to discuss the other two nits while the PR is open..

JAORMX:
> These patches look quite good to me. I have some very minor questions:
> 
>     1. Why is using `github.com/go-logr/logr` directly preferred over using the wrapper over it from `https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/log.go`? This is not so much related to this patchset, but more of FMI.

logr is merely the interface. It's just to make sure that I'm using general functions as opposed to something specific from the implementation.

> 
>     2. Seeing that there used to be the struct `FileIntegrityStatusConditions` made me think if it was meant for extension or shared status between structs that wrap or contain it. Otherwise the substitution seems sane to me.

Well... I'm honestly not sure what the FileIntegrityStatusConditions was meant for. Following Matt's doc, it was more appropriate to have the array directly.
> 
>     3. Do we want to merge patches with WIP in the subject? :)

Fixed.
> 
> 
> Since the tests passed, lgtm can be added, just the WIP in git history would trigger my OCD and also I guess it's easier to discuss the other two nits while the PR is open..

PR #26, opened 2019-12-18
Add logcollector code

JAORMX:
This adds:

* The log collector code. It runs in a loop and uploads configmaps on a
  certain interval (10 min by default).
* The Dockerfile to build the collector
* The necessary changes to the Makefile to build and push the log
collector image
* The dependencies for the collector in our vendor directory

Note that this merely introduces the log collector but doesn't take it into use. Once this merges, we'll need to start creating images (updating the `release` repo) to be able to start using the collector. The actual usage of this will come in a subsequent PR.


cmd/logcollector/main.go
```diff
@@ -0,0 +1,374 @@
...
+	cmd.Flags().String("owner", "", "The compliance scan that owns the configMap objects.")
+	cmd.Flags().String("config-map-prefix", "", "The configMap prefix to upload, typically the podname.")
+	cmd.Flags().String("namespace", "Running pod namespace.", ".")
+	cmd.Flags().Int64("timeout", 600, "How long to wait for the file.")
+	cmd.Flags().Int64("interval", 1800, "every how often does the log collector run.")
+	cmd.Flags().Bool("compress", false, "Always compress the results.")
+}
+
+func getConfigMapName(prefix string) string {
+	return prefix + "-" + os.Getenv("NODE_NAME")
```
jhrozek: You already store NODE_NAME to conf.Name, wouldn't it be more consistent to use it everywhere?

JAORMX: fixed

jhrozek:
oops, I guess my second comment wasn't added because of a race between you pushing the new patches, let me add it again

cmd/logcollector/main.go
```diff
@@ -0,0 +1,374 @@
...
+		// first read (set inode initially)
+		rt.inode = inode
+	} else if rt.inode != inode {
+		// Rotation has happened
+		rt.readoffset = 0
+	}
+}
+
+func updateReadOffset(rt *runtime, file *os.File, contents []byte) error {
+	if _, err := file.Seek(rt.readoffset, 0); err != nil {
```
jhrozek: I guess you can detect shrinking or truncating the file this way, but what about appending? What are even the cases where the file can change?

pkg/common/constants.go
```diff
@@ -5,15 +5,27 @@ const (
...
-	AideScriptPath                 = "/scripts/aide.sh"
-	DaemonSetName                  = "aide-ds"
-	DefaultConfDataKey             = "aide.conf"
-	DefaultConfigMapName           = "aide-conf"
-	FileIntegrityNamespace         = "openshift-file-integrity"
-	LogCollectorDaemonSetName      = "logcollector"
-	OperatorServiceAccountName     = "file-integrity-operator"
-	ReinitDaemonSetName            = "aide-reinit-ds"
+	// IntegrityLogLabelKey tells us that a log was generated by the log collector
+	IntegrityLogLabelKey = "file-integrity.openshift.io/log"
```
jhrozek: This line and some others in the array look like they could use some gofmt

cmd/logcollector/main.go
JAORMX: this would cover appending. Since we're seeking on an offset which keeps growing with the flie itself.

pkg/common/constants.go
JAORMX: make fmt didn't do anything :/

PR #27, opened 2019-12-20
WIP: Take logcollector into use

JAORMX:
This starts taking the log collector into use.

It adds the necessary logic into the reconcile loop of the config map
controller to read the configmaps generated by the log collector and
update the status of the fileintegrity object as needed.


pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -22,7 +24,9 @@ var aideScript = `#!/bin/sh
     while true; do
       echo "running AIDE check.."
       aide -c /tmp/aide.conf
-      echo "AIDE check returned $?.. sleeping"
+      result=$?
+      result > /hostroot/etc/kubernetes/latest-result.log
```
jhrozek: What is this used for?

pkg/common/util.go
```diff
@@ -11,6 +15,47 @@ func IsAideConfig(labels map[string]string) bool {
...
+	_, containsErrorAnnotation := cm.Annotations[IntegrityLogErrorAnnotationKey]
+	_, containsContentKey := cm.Data[IntegrityLogContentKey]
+	return containsErrorAnnotation || !containsContentKey
+}
+
+// IsIntegrityLogAFailure returns whether the given map coming
+// from an integrity check logcollector contains an failure
+func IsIntegrityLogAFailure(cm *corev1.ConfigMap) bool {
+	return cm.Data[IntegrityLogContentKey] != ""
+}
```
jhrozek: Since, Data is a `map[string][string]`. In case a key does not exist in a map, you always get back an empty string if a key does not exist, right? So I'm confused about two things: a) is check for an empty string equivalent to checking with `_, exists`? If yes, it would be at least nice to use the same form here and in `IsIntegrityLogAnError`.
But more importantly, it looks to me like `} else if common.IsIntegrityLogAFailure(cm) {` branch in `handleIntegrityLog` cannot be reached because `IsIntegrityLogAnError` checks for the same condition like `IsIntegrityLogAFailure` in the OR..

pkg/controller/fileintegrity/config_defaults.go
JAORMX: So, currently, we don't know what the result of the scan was, all we know is if there were logs. I was hoping to take this into use.

JAORMX:
@mrogers950 @jhrozek alright folks!

So I wasn't able to finish this, but here's the status.

The logcollector is already in the repo and we already build it as part of the `images` job. This patch takes the log collector into use, by replacing the `pause` container for the logcollector daemon.

Currently the log collector does upload configmaps, and the contoroller will check them... but we need a smarted way to figure out failures. That's why I introduced the writing for the aide check return code  in as part of the aide script. If we take that into account in the logcollector, we'll be able to accurately know if there was a failure, but has been corrected already.

Subsequently, this also appends the statuses. We'll need to remove older statuses from here (and if we want to keep them all, we might wanna introduce a `conditions` section in the status).

Anyway, that's all!

PR #28, opened 2020-01-06
Use logcollector

mrogers950:
continuation of #27


jhrozek:
Oops, I should have noticed the issues when looking at #27 :) The follow up looks good. What were the false positives under `/usr/share`?
Do you have any thought about https://github.com/openshift/file-integrity-operator/pull/27#discussion_r360353030

mrogers950:
@jhrozek It was a lot of stuff from /usr/share/zsh and other directories, perhaps it came from a CoreOS update?
Re: https://github.com/openshift/file-integrity-operator/pull/27#discussion_r360353030 I thought Juan had snuck in an update to those lines after your comment, I will tag the relevant lines in this PR

pkg/common/util.go
```diff
@@ -11,6 +15,47 @@ func IsAideConfig(labels map[string]string) bool {
...
+	_, ok := labels[IntegrityLogLabelKey]
+	return ok
+}
+
+// IsIntegrityLogAnError returns whether the given map coming
+// from an integrity check logcollector contains an error
+func IsIntegrityLogAnError(cm *corev1.ConfigMap) bool {
+	_, containsErrorAnnotation := cm.Annotations[IntegrityLogErrorAnnotationKey]
+	_, containsContentKey := cm.Data[IntegrityLogContentKey]
+	return containsErrorAnnotation || !containsContentKey
```
mrogers950: @jhrozek Re https://github.com/openshift/file-integrity-operator/pull/27#discussion_r360353030 I reviewed this again as I think it was also confusing me..this is a different check since the comma-ok only checks for the presence of the key in the map. Whereas IsIntegrityLogAFailure's form returns true if the key exists and the data is not empty.

jhrozek:
On Wed, Jan 08, 2020 at 03:54:15PM -0800, Matt Rogers wrote:
> @jhrozek It was a lot of stuff from /usr/share/zsh and other directories, perhaps it came from a CoreOS update?

Probably. Now that I can actually install a cluster, I checked what is
there under /usr/share and I wonder if we really want to not track e.g
/usr/share/crypto-policies or /usr/share/ostree/trusted.gpg.d/

mrogers950:
> On Wed, Jan 08, 2020 at 03:54:15PM -0800, Matt Rogers wrote: @jhrozek It was a lot of stuff from /usr/share/zsh and other directories, perhaps it came from a CoreOS update?
> Probably. Now that I can actually install a cluster, I checked what is there under /usr/share and I wonder if we really want to not track e.g /usr/share/crypto-policies or /usr/share/ostree/trusted.gpg.d/

Oh, good point, I'll check those additions again and blacklist the exact ones rather than all of /usr/share.

mrogers950:
> > On Wed, Jan 08, 2020 at 03:54:15PM -0800, Matt Rogers wrote: @jhrozek It was a lot of stuff from /usr/share/zsh and other directories, perhaps it came from a CoreOS update?
> > Probably. Now that I can actually install a cluster, I checked what is there under /usr/share and I wonder if we really want to not track e.g /usr/share/crypto-policies or /usr/share/ostree/trusted.gpg.d/
> 
> Oh, good point, I'll check those additions again and blacklist the exact ones rather than all of /usr/share.

Follow up on this - In reverting the change I never saw this false positive again. So I'm leaving it out of this PR and we can revisit it again if it pops up.

cmd/logcollector/main.go
```diff
@@ -44,10 +46,16 @@ const (
 	crdAPIVersion = "v1alpha1"
 	crdPlurals    = "fileintegrities"
 	maxRetries    = 15
+	// These need to be lessened for normal use.
+	defaultTimeout       = 600
+	defaultInterval      = 1800
+	defaultIndicatorFile = "/hostroot/etc/kubernetes/aide.latest-result.log"
+	uncompressedMaxSize  = 1048570 // Where does this value come from?
```
jhrozek: I think it's roughly the number of bytes in a megabyte (1024**2). That's exactly 1048576 bytes, not sure why here we use 1048570, but the bottom line is that we use 1MB as the threshold for compressing because etcd can only store up to 1MB values.

cmd/logcollector/main.go
```diff
@@ -331,19 +371,39 @@ func doMainLoop(cmd *cobra.Command, args []string) {
...
 
 	for {
+		// Checking the indicator file for the last return code lets us determine if we should collect the aide log.
+		indicatorFile, _ := waitForFile(conf.IndicatorFile, conf.Timeout)
+		if indicatorFile == nil {
+			DBG("Indicator file '%s' doesn't exist, trying again", conf.IndicatorFile)
+			continue
+		}
+
+		checkPassed, err := containsReturnCodeZero(indicatorFile)
```
jhrozek: Could you please add a comment here that `containsReturnCodeZero` closes `indicatorFile`?

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -491,6 +491,10 @@ func aideDaemonset(dsName string, fi *fileintegrityv1alpha1.FileIntegrity) *apps
 								"--config-map-prefix=" + dsName,
 								"--owner=" + fi.Name,
 								"--namespace=" + fi.Namespace,
+								"--timeout=120",
+								"--interval=480",
+								// TODO: remove this for production
+								"--debug=true",
```
jhrozek: Do you want to leave debug in for now?

cmd/logcollector/main.go
```diff
@@ -44,10 +46,16 @@ const (
 	crdAPIVersion = "v1alpha1"
 	crdPlurals    = "fileintegrities"
 	maxRetries    = 15
+	// These need to be lessened for normal use.
```
jhrozek: Why not set the values now to something sensible? Alternatively, we should open a ticket, leaving a comment in the code means we'll just forget (I'm guilty of these comments, too in the compliance-operator..)

pkg/controller/fileintegrity/fileintegrity_controller.go
mrogers950: Yeah for the moment. I'd like to make it so you can turn debugging on through the CRD.

cmd/logcollector/main.go
mrogers950: will do

cmd/logcollector/main.go
mrogers950: Sure, I'll open a ticket.

jhrozek:
Code-wise the current version looks OK, do you know why does CI keep failing?

JAORMX:
Looks good to me! Lets see what CI says.

mrogers950:
got ```error: unable to recognize "/tmp/cluster/insights-live.yaml": Get https://api.ci-op-pk3hlk2v-eb8a9.origin-ci-int-aws.dev.rhcloud.com:6443/api?timeout=32s: dial tcp 34.198.83.53:6443: i/o timeout
Unable to connect to the server: dial tcp 34.198.83.53:6443: i/o timeout
Unable to connect to the server: dial tcp 34.198.83.53:6443: i/o timeout```

/retest

mrogers950:
Another flake:
```Installing from release registry.svc.ci.openshift.org/ci-op-34prnzxd/release@sha256:512e9005c3c02ed88016ccb761a9f6d5a573548b1689f79129d7e69ce856cb63
level=info msg="Credentials loaded from the \"default\" profile in file \"/etc/openshift-installer/.awscred\""
level=warning msg="Found override for release image. Please be warned, this is not advised"
level=info msg="Consuming Install Config from target directory"
level=info msg="Creating infrastructure resources..."
level=error
level=error msg="Error: Error waiting for AMI (ami-032cf936687a75a12) to be ready: timeout while waiting for state to become 'available' (last state: 'pending', timeout: 40m0s)"
level=error
level=error msg="  on ../tmp/openshift-install-849379545/main.tf line 104, in resource \"aws_ami_copy\" \"main\":"
level=error msg=" 104: resource \"aws_ami_copy\" \"main\" {"
level=error
level=error
level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"```

/retest

**Comment by JAORMX** [comment]
timed out :cry:

**Comment by mrogers950** [comment]
Seen in https://github.com/openshift/file-integrity-operator/pull/34
```
            "message": "Failed to pull image \"image-registry.openshift-image-registry.svc:5000/openshift-file-integrity/file-integrity-logcollector:latest\": rpc error: code = Unknown desc = Error reading manifest latest in image-registry.openshift-image-registry.svc:5000/openshift-file-integrity/file-integrity-logcollector: name unknown",
```
We need to have CI build and use logcollector images.

JAORMX:
@mrogers950 that was already implemented in the makefile, with the IMAGE_PATH variable that's set in the `check-if-ci` component. Did something change that it's no longer taken into use?

JAORMX:
Ah, just read this better.

So, `image-registry.openshift-image-registry.svc:5000/openshift-file-integrity/file-integrity-logcollector:latest` is only available when building locally. in CI, that path is not the same, and is exactly what we get in the `check-if-ci` target. We need a way of customizing the path that the logcollector takes to use that path. The tag will be the component that CI built.

PR #29, opened 2020-01-13
images: add aide dockerfiles and make target

mrogers950:

### DIFF
--- a/images/aide/Dockerfile
+++ b/images/aide/Dockerfile
@@ -1,0 +1,2 @@
+FROM centos:centos7
+RUN yum -y install aide && yum clean all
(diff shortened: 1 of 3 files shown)


images/aide/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM centos:centos7
```
jhrozek: Is there a reason to use centos7 and not either centos8 or ubi8? 7.x is rather long in the tooth..
I guess one valid reason might be if the image was supposed to be ran on el7 workers, iirc it is not officially supported to run el8 containers on el7 hosts because the containers might call syscalls the el7 kernel on the host does not know..

PR #33, opened 2020-01-15
Generate CSV for OLM

mrogers950:
Generated using `$ operator-sdk olm-catalog gen-csv --csv-version "0.1.0"`

### DIFF
--- a/deploy/olm-catalog/operator-source.yaml
+++ b/deploy/olm-catalog/operator-source.yaml
@@ -1,0 +1,11 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorSource
+metadata:
+  name: file-integrity-operator
+  namespace: openshift-marketplace
+spec:
+  type: appregistry
+  endpoint: https://quay.io/cnr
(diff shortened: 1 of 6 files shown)


mrogers950:
/hold

Need to fill in a bunch of the fields.

deploy/olm-catalog/operator-source.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorSource
+metadata:
+  name: file-integrity-operator
+  namespace: openshift-marketplace
+spec:
+  type: appregistry
+  endpoint: https://quay.io/cnr
```
JAORMX: might wanna change this.

mrogers950: According to https://github.com/operator-framework/operator-marketplace/blob/master/README.md this is the correct value

JAORMX: OK

PR #36, opened 2020-02-07
Tie AIDE config to FileIntegrity Object

JAORMX:
This ties the AIDE ConfigMap to the FileIntegrity object by giving them
the same name. Effectively, this ends up making the configurations
unique per file-integrity object, as opposed to relying on a single
ConfigMap.

This is the first step on enabling us to use multiple FileIntegrities
per deployment (which would ideally serve different node sets).

Given that the name of the config file matches the name of the
FileIntegrity, we can infer the DaemonSet name from it. Which are also
now unique per CR instance.

### DIFF
--- a/pkg/controller/fileintegrity/fileintegrity_controller.go
+++ b/pkg/controller/fileintegrity/fileintegrity_controller.go
@@ -24,11 +24,6 @@ import (
 
 var log = logf.Log.WithName("controller_fileintegrity")
 
-/**
-* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
-* business logic.  Delete these comments after modifying this file.*
- */
-
 // Add creates a new FileIntegrity Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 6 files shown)


pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -313,26 +311,26 @@ func aideReinitScript() *corev1.ConfigMap {
 
 // reinitAideDaemonset returns a DaemonSet that runs a one-shot pod on each node. This pod touches a file
 // on the host OS that informs the AIDE init container script to back up and reinitialize the AIDE db.
-func reinitAideDaemonset() *appsv1.DaemonSet {
+func reinitAideDaemonset(reinitDaemonSetName string) *appsv1.DaemonSet {
```
mrogers950: We'll need to update aideDaemonset() in the same way, for the config volume ( https://github.com/openshift/file-integrity-operator/blob/c582342eb83b22f8f304f3c37bf0b91d0189217f/pkg/controller/fileintegrity/fileintegrity_controller.go#L529-L528 )

PR #40, opened 2020-02-19
Test newly built workload images in e2e tests

JAORMX:
Currently, we're using the references from the quay image repo to run
the workloads spawn by the compliance-operator in the e2e tests. This is
not ideal since we can't test so easily any changes we do to the
workloads (e.g. resultscollector and resultserver).

This enables that by changing the references to the workloads in the
deploy/operator.yaml, which is what operator-sdk uses to deploy the
operator


mrogers950:
Worked fine for me.
/lgtm

PR #51, opened 2020-03-26
React to expected changes in the cluster

JAORMX:
This takes into account the annotations that the MCO adds to the nodes
in order to react to changes that are "approved" or expected in the
cluster.

Specifically, if the node is expecting changes (currentConfig !=
desiredConfig and the MCD is working) we create a small workload that
creates a file that's read by the AIDE pod. This file tells the AIDE pod
to hold off doing integrity checks.

If the node is done doing changes or it hasn't changed (currentConfig ==
desiredConfig and the MCD is done) we remove the workload that adds the
hold-off file, and create another one that removes that file. This tells
the AIDE pod to continue doing integrity checks as usual.

Big thanks to @ericavonb and @runcom for the help in the algorithm to figure
when the node is going through expected changes.

Co-Authored-By: EvB <evb@redhat.com>
Co-Authored-By: Antonio Murdaca <runcom@linux.com>


JAORMX:
NOTE: This is not ready yet. I'm still testing it out and I still need to get the operator to have the ability to re-init the database on-demand.

Per @mrogers950 currently, the database is only re-initialized if there are config changes. We need a way to re-initialize it on demand in order to let the expected changes take place and not generate false positives.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,311 @@
...
+	if currentConfig != desiredConfig && mcdState == mcfgconst.MachineConfigDaemonStateWorking {
+		// An update is about to take place or already taking place
+		return r.reconcileAddHoldOff(node, reqLogger)
+	} else if currentConfig == desiredConfig && mcdState == mcfgconst.MachineConfigDaemonStateDone {
+		// No update is taking place or it's done already
+		return r.reconcileRemoveHoldOff(node, reqLogger)
+	} else if mcdState == mcfgconst.MachineConfigDaemonStateDegraded {
+		// MCO can't udpate a host, might as well not hold the integrity checks
+		return r.reconcileRemoveHoldOff(node, reqLogger)
+	}
```
JAORMX: @runcom @ericavonb is this logic based on the MCO flags correct?

JAORMX:
This needs https://github.com/openshift/file-integrity-operator/pull/55 before continuing

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	return annotatedFIs
+}
+
+func (r *ReconcileNode) removeHoldoffAnnotationAndReinitFileIntegrityDatabases(fis []*fiv1alpha1.FileIntegrity) error {
+	for _, fi := range fis {
+		fiCopy := fi.DeepCopy()
+		fiCopy.Annotations[common.AideDatabaseReinitAnnotationKey] = ""
+		delete(fiCopy.Annotations, common.IntegrityHoldoffAnnotationKey)
+		if err := r.client.Update(context.TODO(), fiCopy); err != nil {
```
mrogers950: It's ok to not handle an update conflict here? I assume so since the reconcile should pick up where it left off.

JAORMX: @mrogers950 right; if there's a conflict it should just retry. We really do want the database initialization and holdoff removal to take place, and ignoring a conflict would cause those not to be removed.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	// will be empty, and this reconciler just won't do anything. This is fine.
+	if nodeHasMCOAnnotations(node) {
+		fis := r.findRelevantFileIntegrities(node)
+		if isNodeBeingUpdateByMCO(currentConfig, desiredConfig, mcdState) {
+			// An update is about to take place or already taking place
+			if err := r.addHoldOffAnnotations(fis); err != nil {
+				return reconcile.Result{}, err
+			}
+			return r.reconcileAddHoldOff(node, reqLogger)
+		} else if isNodeUpToDateWithMCO(currentConfig, desiredConfig, mcdState) ||
```
jhrozek, line 133: Why only select the uptodate or degraded results? Shouldn't the process resume no matter what once the node is no longer updating?

pkg/controller/node/node_controller.go
jhrozek: FWIW it looks good to me.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		return reconcile.Result{}, err
+	}
+
+	currentConfig := node.Annotations[mcfgconst.CurrentMachineConfigAnnotationKey]
+	desiredConfig := node.Annotations[mcfgconst.DesiredMachineConfigAnnotationKey]
+	mcdState := node.Annotations[mcfgconst.MachineConfigDaemonStateAnnotationKey]
+
+	// NOTE(jaosorior): If, for some reason, the MCO is not running on a deployment, mcdState
+	// will be empty, and this reconciler just won't do anything. This is fine.
+	if nodeHasMCOAnnotations(node) {
```
jhrozek, line 125: What do you think about sprinkling the code with some more logging to see when the holdoff was initiated and in response to what rendered configs?

pkg/controller/node/node_controller.go
JAORMX: Not sure how to detect that to be honest. These seemed like the only phases that were relevant and weren't a catastrophic failure.

pkg/controller/node/node_controller.go
JAORMX: sure. That works for me.

JAORMX:
> In addition to the small comments inline, what do you think about issues that were found prior to an update? Do you think the current model of leaving one failure and one success in the conditions is sufficient? I'm trying to think about a vector when someone would time their attack before an update and make sure the previous issues wouldn't be lost. Or is the expectation that a fileIntegrity failure would trigger some alarm or notification?

I think that in the long run we should move away from adding the node's status in the FileIntegrity object's status. I'm not sure what the ultimate solution is, to be honest...

One option is to have a dedicated node report object. We would then only report on status changes for a specific node:

* When successful, report once
* if it fails after that, report the failure immediately...
* when it keeps failing, don't keep reporting, just keep the first indicated failure.

This "reporting object" could be "FileIntegrityNodeStatus" or something of the sort.

On the other hand, yes, we should also look into adding alerting capabilities & prometheus integration. That could be a nice addition, and who knows, maybe with that the dedicated object would be unnecessary.

pkg/controller/node/node_controller.go
jhrozek: I was wondering about:
``
if updating {
 holdOff()
else if isHoldOff {
 removeHoldOff
}
```
but maybe in the other cases the system is just so fubar that you should reinitialize everything anyway?

jhrozek:
> > In addition to the small comments inline, what do you think about issues that were found prior to an update? Do you think the current model of leaving one failure and one success in the conditions is sufficient? I'm trying to think about a vector when someone would time their attack before an update and make sure the previous issues wouldn't be lost. Or is the expectation that a fileIntegrity failure would trigger some alarm or notification?
> 
> I think that in the long run we should move away from adding the node's status in the FileIntegrity object's status. I'm not sure what the ultimate solution is, to be honest...
> 
> One option is to have a dedicated node report object. We would then only report on status changes for a specific node:
> 
>     * When successful, report once
> 
>     * if it fails after that, report the failure immediately...
> 
>     * when it keeps failing, don't keep reporting, just keep the first indicated failure.
> 
> 
> This "reporting object" could be "FileIntegrityNodeStatus" or something of the sort.
> 
> On the other hand, yes, we should also look into adding alerting capabilities & prometheus integration. That could be a nice addition, and who knows, maybe with that the dedicated object would be unnecessary.

The important part is to not lose the previous failures on reinit. Maybe it could be done in context of the FileIntegrityNodeStatus.
And I should also check how does the reinit work now of course..I haven't actually checked what happens if you have a failure and reinit, does the failure currently go away?

JAORMX:
> > > In addition to the small comments inline, what do you think about issues that were found prior to an update? Do you think the current model of leaving one failure and one success in the conditions is sufficient? I'm trying to think about a vector when someone would time their attack before an update and make sure the previous issues wouldn't be lost. Or is the expectation that a fileIntegrity failure would trigger some alarm or notification?
> > 
> > 
> > I think that in the long run we should move away from adding the node's status in the FileIntegrity object's status. I'm not sure what the ultimate solution is, to be honest...
> > One option is to have a dedicated node report object. We would then only report on status changes for a specific node:
> > ```
> > * When successful, report once
> > 
> > * if it fails after that, report the failure immediately...
> > 
> > * when it keeps failing, don't keep reporting, just keep the first indicated failure.
> > ```
> > 
> > 
> > This "reporting object" could be "FileIntegrityNodeStatus" or something of the sort.
> > On the other hand, yes, we should also look into adding alerting capabilities & prometheus integration. That could be a nice addition, and who knows, maybe with that the dedicated object would be unnecessary.
> 
> The important part is to not lose the previous failures on reinit. Maybe it could be done in context of the FileIntegrityNodeStatus.
> And I should also check how does the reinit work now of course..I haven't actually checked what happens if you have a failure and reinit, does the failure currently go away?

It doesn't go away. Only the previous "success" goes away. The failure stays. The only way to identify which is the current state is to check the timestamps.

jhrozek:
> > > > In addition to the small comments inline, what do you think about issues that were found prior to an update? Do you think the current model of leaving one failure and one success in the conditions is sufficient? I'm trying to think about a vector when someone would time their attack before an update and make sure the previous issues wouldn't be lost. Or is the expectation that a fileIntegrity failure would trigger some alarm or notification?
> > > 
> > > 
> > > I think that in the long run we should move away from adding the node's status in the FileIntegrity object's status. I'm not sure what the ultimate solution is, to be honest...
> > > One option is to have a dedicated node report object. We would then only report on status changes for a specific node:
> > > ```
> > > * When successful, report once
> > > 
> > > * if it fails after that, report the failure immediately...
> > > 
> > > * when it keeps failing, don't keep reporting, just keep the first indicated failure.
> > > ```
> > > 
> > > 
> > > This "reporting object" could be "FileIntegrityNodeStatus" or something of the sort.
> > > On the other hand, yes, we should also look into adding alerting capabilities & prometheus integration. That could be a nice addition, and who knows, maybe with that the dedicated object would be unnecessary.
> > 
> > 
> > The important part is to not lose the previous failures on reinit. Maybe it could be done in context of the FileIntegrityNodeStatus.
> > And I should also check how does the reinit work now of course..I haven't actually checked what happens if you have a failure and reinit, does the failure currently go away?
> 
> It doesn't go away. Only the previous "success" goes away. The failure stays. The only way to identify which is the current state is to check the timestamps.

OK, then I think even the current state is fine.

PR #52, opened 2020-03-26
Add numerical summary how many files were added, changed or removed

jhrozek:
Adds three new fields to the failed conditions - filesAdded,
filesChanged, filesRemoved. The number of files is parsed out from the
aide output and passed to the configmap controller by annotating the
configmaps.


jhrozek:
I marked this as WIP as I'm having issues with running the e2e tests locally

jhrozek:
OK, looks like the e2e tests passed here. I'm going to try again tomorrow locally.

@mrogers950 could you please take a look at the code? I'm mostly wondering about adding more tests. Initially I wanted to test all three cases (add/remove/change) with different number of changes to protect against e.g. us copy-pasting the wrong annotation or something. But while changes are trivial, we could change files like we do now by appending a comment, additions might also be sort-of trivial, we could just drop a file with random filename, deletions are hard if we want to make the tests somehow idempotent. Maybe we could mv a file and then have a fixture that mv-es it back? What would be your preference for a good test?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -266,9 +272,10 @@ func (r *ReconcileConfigMap) updateStatus(fi *fileintegrityv1alpha1.FileIntegrit
 func getConfigMapForFailureLog(cm *corev1.ConfigMap) *corev1.ConfigMap {
 	failedCM := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      cm.Name + "-failed",
-			Namespace: cm.Namespace,
-			Labels:    cm.Labels,
+			Name:        cm.Name + "-failed",
+			Namespace:   cm.Namespace,
+			Labels:      cm.Labels,
+			Annotations: cm.Annotations,
```
JAORMX: you could have done:
```
ObjectMeta: cm.ObjectMeta,
```

PR #53, opened 2020-03-26
Remove unused functions

jhrozek:


jhrozek:
@mrogers950 was there a reason these functions were kept around, maybe for easier debugging? If yes, it would be fine to keep them around, but maybe with a comment.

PR #56, opened 2020-03-31
reinit e2e: Test that nodes are in good state after re-init

JAORMX:
The re-init e2e test was only verifying that the FileIntegrity is
Active. But this is not enough. We also need to verify that the nodes
are in a good state (Succeeded) after the database has been
re-initialized.

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	goctx "context"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


tests/e2e/helpers.go
```diff
@@ -474,6 +475,30 @@ func waitForFailedStatusForNode(t *testing.T, f *framework.Framework, namespace,
...
+	wait.Poll(interval, timeout, func() (bool, error) {
+		getErr := f.Client.Get(goctx.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, fi)
+		if getErr != nil {
+			t.Logf("Retrying. Got error: %v\n", getErr)
+			lastErr = getErr
+			return false, nil
+		}
+
+		for _, status := range fi.Status.Statuses {
+			if status.Condition != fileintv1alpha1.NodeConditionSucceeded {
```
mrogers950: @JAORMX  Ah, this check will always fail.. when a node has a fail and then later a succeed, the latest failed entry is still left. But it will have a succeeded condition that is newer than the last failure, that's what you would want to check for here..

JAORMX: I see. Thank for clarifying! This sounds like a bug though

JAORMX: Also, the problem right now is that this is failing in the first integrity check. So, in that case, there isn't any other condition.

JAORMX: Got it! I understand now why this works this way. Perhaps it would be better to instead put the latest results first, so it's easier to iterate. But that can be done in a subsequent PR.

JAORMX: OK, I changed the code and it seems it still is only in failed state.

PR #59, opened 2020-04-22
FI controller: Make the interval and the debug flag configurable

jhrozek:
FI ctrl: Make debugging of the aggregator configurable
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

The aggregator pod can optionally produce some debug information, but the
debug flag was hardcoded to false. This patch extends the FileIntegritySpec
struture with a new field, debug, that defaults to false. The admin can
enable debugging of the logcollector by setting the debug flag to true.

FI ctrl: Don't set the timeout explicitly
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Setting the timeout explicitly or even making the timeout configurable is 
not really needed, because if the timeout occurs, the loop that waits for
the files the aggregator is watching simply takes another iteration.

FI ctrl: Make the interval configurable
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Extends the FileIntegrityConfig structure with a new member Interval. If 
not set, its value defaults to 10 seconds, otherwise the configured value
is used for both the aide integrity scanner and the interval in which the
logcollector checks for aide results.

Jira: CMP-436

### DIFF
--- a/cmd/manager/logcollector.go
+++ b/cmd/manager/logcollector.go
@@ -49,7 +49,6 @@ const (
 	maxRetries    = 15
 	// These need to be lessened for normal use.
 	defaultTimeout       = 600
-	defaultInterval      = 1800
 	defaultIndicatorFile = "/hostroot/etc/kubernetes/aide.latest-result.log"
 	uncompressedMaxSize  = 1048570 // 1MB for etcd limit
 )
@@ -109,7 +108,7 @@ func defineFlags(cmd *cobra.Command) {
 	cmd.Flags().String("config-map-prefix", "", "Prefix for the configMap name, typically the podname.")
 	cmd.Flags().String("namespace", "Running pod namespace.", ".")
 	cmd.Flags().Int64("timeout", defaultTimeout, "How long to poll for the log and indicator files in seconds.")
-	cmd.Flags().Int64("interval", defaultInterval, "How often to recheck for .")
+	cmd.Flags().Int64("interval", common.DefaultGracePeriod, "How often to recheck for AIDE results.")
 	cmd.Flags().Bool("compress", false, "Use gzip+base64 to compress the log file contents.")
 	cmd.Flags().Bool("debug", false, "Print debug messages")
 }
--- a/deploy/crds/file-integrity.openshift.io_fileintegrities_crd.yaml
+++ b/deploy/crds/file-integrity.openshift.io_fileintegrities_crd.yaml
@@ -45,6 +45,8 @@ spec:
                 namespace:
                   type: string
               type: object
+            debug:
+              type: boolean
             nodeSelector:
               additionalProperties:
                 type: string
--- a/deploy/crds/file-integrity.openshift.io_fileintegrities_crd.yaml
+++ b/deploy/crds/file-integrity.openshift.io_fileintegrities_crd.yaml
@@ -35,6 +35,9 @@ spec:
               description: FileIntegrityConfig defines the name, namespace, and data
                 key for an AIDE config to use for integrity checking.
               properties:
+                gracePeriod:
+                  description: Time between individual aide scans
+                  type: integer
                 key:
                   type: string
                 name:
--- a/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
+++ b/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
@@ -28,6 +28,7 @@ const (
 type FileIntegritySpec struct {
 	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
 	Config       FileIntegrityConfig `json:"config"`
+	Debug        bool                `json:"debug,omitempty"`
 }
 
 // FileIntegrityConfig defines the name, namespace, and data key for an AIDE config to use for integrity checking.
--- a/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
+++ b/pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
@@ -36,6 +36,8 @@ type FileIntegrityConfig struct {
 	Name      string `json:"name,omitempty"`
 	Namespace string `json:"namespace,omitempty"`
 	Key       string `json:"key,omitempty"`
+	// Time between individual aide scans
+	GracePeriod int `json:"gracePeriod,omitempty"`
 }
 
 // FileIntegrityStatus defines the observed state of FileIntegrity
--- a/pkg/common/util.go
+++ b/pkg/common/util.go
@@ -2,11 +2,10 @@ package common
 
 import (
 	"fmt"
-	"os"
-
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kerr "k8s.io/apimachinery/pkg/api/errors"
+	"os"
 )
 
 type FileIntegrityComponent uint
@@ -40,6 +39,13 @@ func GetComponentImage(component FileIntegrityComponent) string {
 // label that indicates that this is an AIDE config.
 func IsAideConfig(labels map[string]string) bool {
 	_, ok := labels[AideConfigLabelKey]
+	return ok
+}
+
+// IsAideScript returns whether the given map contains a
+// label that indicates that this is an AIDE script
+func IsAideScript(labels map[string]string) bool {
+	_, ok := labels[AideScriptLabelKey]
 	return ok
 }
 
@@ -99,6 +105,12 @@ func IgnoreAlreadyExists(err error) error {
 		return nil
 	}
 	return err
+}
+
+// GetScriptName returns the name of a configMap for a FI object with a
+// given name
+func GetScriptName(fiName string) string {
+	return AideScriptConfigMapPrefix + "-" + fiName
 }
 
 // GetDaemonSetName gets the name of the owner (usually a FileIntegrity CR) and
--- a/pkg/controller/configmap/configmap_controller.go
+++ b/pkg/controller/configmap/configmap_controller.go
@@ -103,6 +103,8 @@ func (r *ReconcileConfigMap) Reconcile(request reconcile.Request) (reconcile.Res
 
 	if common.IsAideConfig(instance.Labels) {
 		return r.reconcileAideConf(instance, reqLogger)
+	} else if common.IsAideScript(instance.Labels) {
+		return r.reconcileAideScript(instance, reqLogger)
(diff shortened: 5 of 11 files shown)


jhrozek:
@mrogers950: There are no tests in this PR. The naive thing to do would be to check the CM and the args of the logcollector container. A more involve test might set a largish interval (120s?) and make sure that the file change is not detected before, maybe at a 100s mark or so. But I'm not sure how reliable this test would be. Do you have some suggestions?

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -36,6 +37,7 @@ type FileIntegrityConfig struct {
 	Name      string `json:"name,omitempty"`
 	Namespace string `json:"namespace,omitempty"`
 	Key       string `json:"key,omitempty"`
+	Interval  int    `json:"interval,omitempty"`
```
mrogers950, line 40: I wonder if there is a different wording we can use to convey that it would be the length of the pause between scans, not that a scan will be done every Interval seconds.. Thinking something like GracePeriod

mrogers950:
> A more involve test might set a largish interval (120s?) and make sure that the file change is not detected before, maybe at a 100s mark or so. But I'm not sure how reliable this test would be. Do you have some suggestions?

That test might be fine... but it would be hard to gauge what the buffer zone should be to ensure it worked, and could flake due to other things. It would probably be better off to just confirm in the configmap that the setting made its way into the script, no need to test if the actual sleep command worked.

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
JAORMX: When in doubt, it's nice to resort to previous art.

AIDE rechecks are normally configured via cronjobs. So admins would be more acquainted with that notation instead. So I would suggest using that. And... for that, I would suggest the usage of "schedule" as with CronJobs: https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

JAORMX: If you're not too keen on using a cronjob-like notation, in this case, I am a bit keener on "interval" since grace period to me suggests that it's a vulnerable period or a period where we don't care about the check, which it isn't.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -546,11 +557,8 @@ func aideDaemonset(dsName string, fi *fileintegrityv1alpha1.FileIntegrity) *apps
 								"--file=" + aideLogPath,
 								"--config-map-prefix=" + dsName,
 								"--owner=" + fi.Name,
-								"--namespace=" + fi.Namespace,
-								"--timeout=2",
-								"--interval=10",
-								// TODO: remove this for production
-								"--debug=true",
+								"--namespace=" + fi.Namespace, "--interval=" + strconv.Itoa(interval),
+								"--debug=" + strconv.FormatBool(fi.Spec.Debug),
```
JAORMX, line 561: what happened to the timeout?

jhrozek: See the commit with subject ` FI ctrl: Don't set the timeout explicitly`. Basically it wasn't really needed to be set to anything non-default, so it's better to have one less parameter.

jhrozek:
New patches were pushed that allow the grace period to be set on the fly, fix some logic bugs and add a e2e test.

deploy/crds/file-integrity.openshift.io_fileintegrities_crd.yaml
```diff
@@ -35,19 +35,25 @@ spec:
...
               type: object
+            debug:
+              type: boolean
             nodeSelector:
               additionalProperties:
                 type: string
               type: object
           required:
           - config
+          - debug
```
JAORMX, line 56: why is debug required? it could easily be optional.

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -28,6 +28,7 @@ const (
 type FileIntegritySpec struct {
 	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
 	Config       FileIntegrityConfig `json:"config"`
+	Debug        bool                `json:"debug"`
```
JAORMX, line 31: let's add omitempty so Debug doesn't become a required parameter.

pkg/common/util.go
```diff
@@ -112,3 +116,51 @@ func GetDaemonSetName(name string) string {
 func GetReinitDaemonSetName(name string) string {
 	return ReinitDaemonSetPrefix + "-" + name
 }
+
+func RestartFileIntegrityDs(c client.Client, baseName string, logger logr.Logger) error {
```
JAORMX, line 120: Let's add docs to this. It can easily be confused with the ReinitDS

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -459,6 +507,11 @@ func aideDaemonset(dsName string, fi *fileintegrityv1alpha1.FileIntegrity) *apps
 	runAs := int64(0)
 	mode := int32(0744)
 
+	gracePeriod := fi.Spec.Config.GracePeriod
+	if gracePeriod < 10 {
+		gracePeriod = 10
+	}
```
JAORMX, line 513: why do we set the bottom limit to 10 here, and above we use the `DefaultGracePeriod` constant? let's use that same constant for both the script and the log collector.

jhrozek: The bottom limit is set to 10 to not overwhelm the nodes with almost a tight loop of aide scans. It's also the previous default, so the changes just doesn't allow to set a tighter loop than the previous hardcoded defaults.
I've used the same constant for the collector and the script now.

pkg/common/util.go
jhrozek: The latest version no longer splits out the function to the common module, but I added the docstring nonetheless.

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
jhrozek: fixed

jhrozek:
This version has mainly these changes:
 - the DS is no longer restarted by the FI controller, but by the CM controller, following pretty much the same logic as changing the config does
- the aide script is no longer a singleton CM used by all DS-es, but since the grace period can be set per FI instance, also the scripts are now created per FI instance.

jhrozek:
Rebased atop the PR that folds logcollector into operator.

cmd/logcollector/main.go
```diff
@@ -18,4 +18,4 @@ limitations under the License.
 package main
 
 func main() {
-}
+}
```
mrogers950, line 21: editor bug? :)

cmd/manager/logcollector.go
```diff
@@ -109,7 +108,7 @@ func defineFlags(cmd *cobra.Command) {
 	cmd.Flags().String("config-map-prefix", "", "Prefix for the configMap name, typically the podname.")
 	cmd.Flags().String("namespace", "Running pod namespace.", ".")
 	cmd.Flags().Int64("timeout", defaultTimeout, "How long to poll for the log and indicator files in seconds.")
-	cmd.Flags().Int64("interval", defaultInterval, "How often to recheck for .")
+	cmd.Flags().Int64("interval", common.DefaultGracePeriod, "How often to recheck for .")
```
mrogers950, line 111: I never noticed that I left the description text incomplete. Can you fill it in?

pkg/controller/configmap/configmap_controller.go
```diff
@@ -12,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
```
mrogers950, line 15: Lets move this back

pkg/controller/configmap/configmap_controller.go
```diff
@@ -142,20 +144,43 @@ func (r *ReconcileConfigMap) reconcileAideConf(instance *corev1.ConfigMap, logge
...
+	err = restartFileIntegrityDs(r.client, common.GetDaemonSetName(instance.Name))
 	if err != nil {
-		logger.Error(err, "error getting daemonSet")
+		logger.Error(err, "error restarting daemonSet")
 		return reconcile.Result{}, err
 	}
 
-	if err := deleteDaemonSetPods(r.client, ds); err != nil {
-		logger.Error(err, "error deleting daemonSet pods")
+	logger.Info("DBG: rollout triggered, clearing update annotation")
```
mrogers950, line 153: I think we can remove this debug line from long ago, it's not particularly helpful/accurate.

tests/e2e/helpers.go
```diff
@@ -3,6 +3,7 @@ package e2e
 import (
 	goctx "context"
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
mrogers950, line 6: nit: move down to another import block

mrogers950:
Awesome work, Just a couple nits and I'm all good for merge.

cmd/logcollector/main.go
jhrozek: No, Punk-rock themed rebasing..

cmd/manager/logcollector.go
jhrozek: ok, done

pkg/controller/configmap/configmap_controller.go
jhrozek: done

pkg/controller/configmap/configmap_controller.go
jhrozek: done

tests/e2e/helpers.go
jhrozek: done

jhrozek:
One more push to resolve a rebase conflict

jhrozek:
@JAORMX @mrogers950 could either of you add a lgtm so the patch gets merged? (Or can be merged manually?)

PR #60, opened 2020-04-22
Node controller: Several fixes

jhrozek:
Please see the commit messages for more details, mostly this PR makes the
holdoff pods code work as it makes the configmap names consistent across the
code and fixes some other issues with the pod or the scripts they use.

In addition, I wonder if the controller shouldn't remove the pods when they
are done? Also I guess tests should be improved since they didn't catch
the issues the firts time around?

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -307,7 +307,7 @@ func newGenericHoldOffIntegrityCheckPod(holdoffCmName, holdoffScriptName string,
 	priv := true
 	runAs := int64(0)
 	mode := int32(0744)
-	labels := map[string]string{
+	podLabels := map[string]string{
 		common.IntegrityPodLabelKey: "",
 	}
 
@@ -315,7 +315,7 @@ func newGenericHoldOffIntegrityCheckPod(holdoffCmName, holdoffScriptName string,
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      holdoffCmName,
 			Namespace: common.FileIntegrityNamespace,
-			Labels:    labels,
+			Labels:    podLabels,
 		},
 		Spec: corev1.PodSpec{
 			// Schedule directly to the node (skip the scheduler)
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -167,21 +167,23 @@ func (r *ReconcileNode) reconcileRemoveHoldOff(node *corev1.Node, reqLogger logr
 
 func (r *ReconcileNode) reconcileCreateWorkloadForNode(name, script string, node *corev1.Node, reqLogger logr.Logger) (reconcile.Result, error) {
 	cmToBeCreated := newGenericHoldOffCM(name, script, node)
+	reqLogger.Info("Creating CM", "name", cmToBeCreated)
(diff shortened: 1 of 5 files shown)


mrogers950:
The fixes appear to make sense, but it's really strange that we didn't see testing issues from this.

pkg/controller/node/node_controller.go
```diff
@@ -301,16 +303,17 @@ func newGenericHoldOffCM(name, script string, node *corev1.Node) *corev1.ConfigM
...
 	priv := true
 	runAs := int64(0)
 	mode := int32(0744)
 	labels := map[string]string{
 		common.IntegrityPodLabelKey: "",
 	}
+
 	return &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      node.Name + "-" + holdoffScriptName,
```
JAORMX, line 315: this shouldn't change, as the pod is unique to a node. Else we'll only have one pod, and subsequent holdoff operatiohns won't work.

JAORMX:
> Please see the commit messages for more details, mostly this PR makes the
> holdoff pods code work as it makes the configmap names consistent across the
> code and fixes some other issues with the pod or the scripts they use.
> 
> In addition, I wonder if the controller shouldn't remove the pods when they
> are done?

That's a possibility but not strictly needed. cleanup is done when a toggle happens (from holdoff to it being cancelled, and so on.

> Also I guess tests should be improved since they didn't catch
> the issues the firts time around?

Yep, I'm very confused as to why the e2e tests didn't catch this.

pkg/controller/node/node_controller.go
jhrozek: The resulting string is essentially the same. `holdoffCmName` is already per node:
```
func newGenericHoldOffCM(name, script string, node *corev1.Node) *corev1.ConfigMap {
	return &corev1.ConfigMap{
		ObjectMeta: metav1.ObjectMeta{
			Name:      node.Name + "-" + name,
			Namespace: common.FileIntegrityNamespace,
			Labels: map[string]string{
				common.IntegrityCMLabelKey: "",
			},
		},
		Data: map[string]string{
			name: script,
		},
	}
}
```
That was actually my next question: Do we need the CM to be per-node? They are the same, can't we have a single CM?

JAORMX: the CM doesn't need to be per node. that was a copy-pasta issue...

jhrozek:
I'm marking this as WIP because some work is still needed (namely the test now takes too long), but hopefully it should be clearer why the tests didn't catch the issue. Plus, some more tidyups are needed like making the CM a singleton. Nonetheless, comments are welcome.

jhrozek:
Pushed the current state and removed WIP as the PR was blocking Matt.

jhrozek:
Looks like the last patch in the series was not completely working, I backed it off until it work 100%

jhrozek:
```
error: unable to read image default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com/ci-op-knh1bkh9/stable@sha256:0803f024975475572280566ebe9fe5552d130fefbc790cc5c194bc98d656f0ec: received unexpected HTTP status: 502 Bad Gateway
```

mrogers950:
All green! Merging.

PR #65, opened 2020-05-01
Fix several race conditions in the FI operator

jhrozek:
Please check out the commit messages for more details.

### DIFF
--- a/pkg/common/util.go
+++ b/pkg/common/util.go
@@ -84,7 +84,7 @@ func GetConfigMapNodeName(cm *corev1.ConfigMap) (string, error) {
 }
 
 func DaemonSetIsReady(ds *appsv1.DaemonSet) bool {
-	return ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable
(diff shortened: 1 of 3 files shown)


pkg/common/util.go
```diff
@@ -84,7 +84,7 @@ func GetConfigMapNodeName(cm *corev1.ConfigMap) (string, error) {
 }
 
 func DaemonSetIsReady(ds *appsv1.DaemonSet) bool {
-	return ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable
+	return ds.Status.DesiredNumberScheduled > 0 && ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable
```
jhrozek, line 87: So I'm not sure about this change myself, because this would never remove a mislabeled DS (one that uses a nodeSelector that does not match any nodes). But I couldn't come up with any other sensible fix.

JAORMX: This sounds like the best solution so far. At least it seems to solve the issue.

PR #66, opened 2020-05-04
FI: Don't use a third-party container image

jhrozek:
Using container images from a public registry migth break disconnected
installs.

### DIFF
--- a/pkg/controller/fileintegrity/fileintegrity_controller.go
+++ b/pkg/controller/fileintegrity/fileintegrity_controller.go
@@ -110,6 +110,19 @@ func (r *ReconcileFileIntegrity) handleDefaultConfigMaps(f *fileintegrityv1alpha
 	}
 
 	if err := r.client.Get(context.TODO(), types.NamespacedName{
+		Name:      common.PauseConfigMapName,
+		Namespace: common.FileIntegrityNamespace,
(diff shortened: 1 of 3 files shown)


pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -423,8 +448,15 @@ func reinitAideDaemonset(reinitDaemonSetName string, fi *fileintegrityv1alpha1.F
 					// make this an endless loop
 					Containers: []corev1.Container{
 						{
-							Name:  "pause",
-							Image: "gcr.io/google_containers/pause",
+							Name:    "pause",
+							Command: []string{common.PausePath},
+							Image:   "registry.access.redhat.com/ubi8/ubi-minimal:latest",
```
mrogers950, line 453: Wouldn't this image also be a problem in a disconnected install?
Perhaps it's bad form, but could we call the script from a shell using an image we already have (that is based on ubi8), like the operator image itself?

jhrozek: To be honest, I don't know what is allowed and what is not. But using an image we already have is probably the best course of action.
Thank you for the review, I'll change the patch.

PR #67, opened 2020-05-08
make: Fix make run after upgrade to operator-sdk 1.17

jhrozek:
Otherwise we'd just receive:
INFO[0000] --namespace is deprecated; use --watch-namespace instead.
and the operator then quits

In addition to the --watch-namespace change, there is also another
change that adds "operator-flags operator" which is needed since we
collapsed all the functionality into a single binary.


jhrozek:
```
 panic: test timed out after 30m0s 
```
let's try this again..

mrogers950:
@jhrozek I'm addressing this in https://github.com/openshift/file-integrity-operator/pull/82

PR #72, opened 2020-06-15
Revised AIDE daemonSet

mrogers950:
This PR makes for a simpler AIDE daemonSet by combining the logcollector, aide check, and re-init functionality into a single go daemon. Aside from the needed e2e changes, I revised some logging statements in order to give a clearer flow between controllers particularly during re-initialization, as well as adding code to handle configuration changes from an existing FI  (see updateDaemonSetPodArgs()).

### DIFF
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -1,0 +1,376 @@
+/*
+Copyright © 2020 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"sync"
+	"syscall"
+	"time"
+
+	"k8s.io/api/events/v1beta1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/file-integrity-operator/pkg/common"
+
+	"github.com/spf13/cobra"
+
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	aideDBPath       = "/hostroot/etc/kubernetes/aide.db.gz"
+	aideLogPath      = "/hostroot/etc/kubernetes/aide.log"
+	backupTimeFormat = "20060102T15_04_05"
+	aideReinitFile   = "/hostroot/etc/kubernetes/aide.reinit"
+	aideHoldoffFile  = "/hostroot/etc/kubernetes/holdoff"
+)
+
+var daemonCmd = &cobra.Command{
+	Use:   "daemon",
+	Short: "daemon",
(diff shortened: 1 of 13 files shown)


cmd/manager/daemon.go
```diff
@@ -0,0 +1,307 @@
...
+			}
+			_, err = f.WriteString("\n")
+			if err != nil {
+				LOG(err.Error())
+				f.Close()
+				time.Sleep(time.Second)
+				continue
+			}
+			f.Sync()
+			f.Close()
```
JAORMX, line 287: we should probably read the error here and react appropriately. Given that this is all concurrent, there might still be some synchronization issues.

tests/e2e/helpers.go
```diff
@@ -82,6 +82,7 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/kubernetes/manifests
 !/hostroot/etc/docker/certs.d
 !/hostroot/etc/selinux/targeted
+!/hostroot/etc/openvswitch/conf.db
```
JAORMX, line 85: Might want to add this in a separate commit.

cmd/manager/daemon.go
```diff
@@ -0,0 +1,342 @@
...
+// The holdoff file is the signal from the node controller to pause the aide scan.
+// We do not make this pause the logCollector loop, and we might want to.
+func holdOffLoop(rt *daemonRuntime, conf *daemonConfig) {
+	for {
+		_, statErr := os.Stat(aideHoldoffFile)
+		if statErr == nil {
+			rt.SetHolding("holdOffLoop", true)
+		} else if os.IsNotExist(statErr) {
+			rt.SetHolding("holdOffLoop", false)
+		}
```
JAORMX, line 241: What do we do if we have another error? e.g. any other error mentioned in the stat man page: https://linux.die.net/man/3/stat

We should at least log that, but it should also be handled somehow (logging and a FIXME comment would be sufficient for this PR).

cmd/manager/daemon.go
```diff
@@ -0,0 +1,342 @@
...
+			if err := runAideInitDBCmd(); err != nil {
+				LOG(err.Error())
+				time.Sleep(time.Second)
+				continue
+			}
+
+			if err := removeAideReinitFile(); err != nil {
+				LOG(err.Error())
+				time.Sleep(time.Second)
+				continue
```
JAORMX, line 289: if we can't remove the re-init file we should at least issue an event so the admin is aware of the issue. It could be that there is a funky permissions error and we can' remove the file.

cmd/manager/daemon.go
```diff
@@ -0,0 +1,342 @@
...
+			LOG("initialization finished")
+			rt.UnlockAideFiles("reinitLoop")
+		} else if initStatErr == nil {
+			rt.SetInitializing("reinitLoop", true)
+			rt.LockAideFiles("reinitLoop")
+			LOG("re-initializing aide")
+
+			if err := backUpAideFiles(); err != nil {
+				LOG(err.Error())
+				time.Sleep(time.Second)
```
JAORMX, line 270: same here, if we can't back-up the file we should at least try to issue an event... would this be a fatal error though? seems to me like the errors that could come from `os.Rename` are quite fatal IMO https://man7.org/linux/man-pages/man2/rename.2.html

cmd/manager/daemon.go
```diff
@@ -0,0 +1,342 @@
...
+				continue
+			}
+
+			if err := initAideLog(); err != nil {
+				LOG(err.Error())
+				time.Sleep(time.Second)
```
JAORMX, line 276: This should be a fatal error too IMO. One of the issues that could happen here is that the process runs out of file descriptors if it's running for a long time (or they're limited inside the container), restarting the container could potentially fix that.

PR #75, opened 2020-06-16
Remove duplicates CRDs

pichuang:
Fixed the issue on v0.1.2 and v0.1.3

Signed-off-by: Phil Huang <phil.huang@redhat.com>


mrogers950:
@pichuang thank you for your patch, however our CSV files are auto-generated and pull from the files we have in deploy/crds. We had some duplicate CRD definitions there that I removed in https://github.com/openshift/file-integrity-operator/pull/71.
After we merge https://github.com/openshift/file-integrity-operator/pull/74 , using 0.1.4 should not have any issues with duplicate CRDs.

That said I'm not sure if we should retroactively change these manifests. Since we're not officially releasing the old versions we are mostly keeping them around for record. @jhrozek @JAORMX thoughts?

pichuang:
@mrogers950 Got it. It's fine for me if you guys need to close the ticket. : )

PR #84, opened 2020-07-03
daemon: Change result handling to use channels instead of mutexes

JAORMX:
This reduces the code a little bit and makes the log collector and aide
loop have their synchronization dependent on whether there's a result or
not.

This way, the only "sleep" is done in the aide loop, and the log
collector will only collect if there is indeed a result ready.


JAORMX:
@mrogers950 what do you think?

mrogers950:
Yay!
/lgtm

PR #86, opened 2020-07-20
Enable tolerations for fileIntegrity

mrogers950:
Add a `tolerations` field to FileIntegritySpec, allowing running on nodes with
custom taints. Defaults to master node tolerations.

### DIFF
--- a/deploy/crds/fileintegrity.openshift.io_fileintegrities_crd.yaml
+++ b/deploy/crds/fileintegrity.openshift.io_fileintegrities_crd.yaml
@@ -51,6 +51,51 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              tolerations:
+                default:
+                - effect: NoSchedule
+                  key: node-role.kubernetes.io/master
+                  operator: Exists
+                description: Specifies tolerations for custom taints. Defaults to
+                  allowing scheduling on master nodes.
+                items:
+                  description: The pod this Toleration is attached to tolerates any
+                    taint that matches the triple <key,value,effect> using the matching
+                    operator <operator>.
+                  properties:
+                    effect:
+                      description: Effect indicates the taint effect to match. Empty
+                        means match all taint effects. When specified, allowed values
+                        are NoSchedule, PreferNoSchedule and NoExecute.
+                      type: string
+                    key:
+                      description: Key is the taint key that the toleration applies
+                        to. Empty means match all taint keys. If the key is empty,
+                        operator must be Exists; this combination means to match all
+                        values and all keys.
+                      type: string
+                    operator:
+                      description: Operator represents a key's relationship to the
+                        value. Valid operators are Exists and Equal. Defaults to Equal.
+                        Exists is equivalent to wildcard for value, so that a pod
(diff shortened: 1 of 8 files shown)


mrogers950:
/hold
e2e-aws didn't run the tests, but reported passed..

mrogers950:
@jhrozek tests are fixed now. PTAL

deploy/crds/fileintegrity.openshift.io_fileintegrities_crd.yaml
```diff
@@ -51,6 +51,51 @@ spec:
                 additionalProperties:
                   type: string
                 type: object
+              tolerations:
+                default:
+                - effect: NoSchedule
```
jhrozek, line 56: Do you think the tolerations are worth mentioning in the README?

tests/e2e/helpers.go
```diff
@@ -298,6 +300,96 @@ func getNumberOfWorkerNodes(c kubernetes.Interface) (int, error) {
...
+			NodeSelector: map[string]string{
+				// Schedule on the tainted host
+				corev1.LabelHostname: taintedNode.Labels[corev1.LabelHostname],
+			},
+			Config: fileintv1alpha1.FileIntegrityConfig{},
+			Tolerations: []corev1.Toleration{
+				{
+					Key:      taintKey,
+					Operator: corev1.TolerationOpExists,
+					Effect:   corev1.TaintEffectNoSchedule,
```
jhrozek, line 346: Just a question: This combination of taitns and tolerations means that the scheduler will *only* be able to schedule the FI DS on a pod tainted with the corresponding taint? (The NoSchedule reads strange to me, it reads like "Don't schedule", OTOH I've only today read again about taints and tolerations...)

tests/e2e/helpers.go
```diff
@@ -298,6 +300,96 @@ func getNumberOfWorkerNodes(c kubernetes.Interface) (int, error) {
...
+	})
+	if pollErr != nil {
+		t.Errorf("error confirming DS replica amount: (%v) (%v)", pollErr, lastErr)
+	}
+
+	// wait to go active
+	err = waitForScanStatus(t, f, namespace, testIntegrityName, fileintv1alpha1.PhaseActive)
+	if err != nil {
+		t.Error("Timed out waiting for scan status to go Active")
+	}
```
jhrozek, line 389: The test itself also waits for the scan to go active, are both needed?

jhrozek:
Looks good! Left some questions inline.

tests/e2e/helpers.go
mrogers950: With that toleration it can schedule on non-tainted nodes, and also nodes that have the taint of `NoSchedule`. I also always found it misleading, but from the toleration perspective it's just a label to match.

deploy/crds/fileintegrity.openshift.io_fileintegrities_crd.yaml
mrogers950: Yep, I'll update the README

PR #92, opened 2020-08-10
Revert "debug sleeps"

jhrozek:
This reverts commit cfdf1b129dabf956dce72553e42d5e6c73cae007.which
appears to be commited by accident.


JAORMX:
/lgtm

oops!

mrogers950:
There's a race with the status that I was looking into with https://github.com/openshift/file-integrity-operator/pull/91 that was making the tests fail, I just forgot to hold the PR. :)

We can close this and I'll remove the sleeps with the PR for the fix.

PR #95, opened 2020-08-17
Move OK configMap log to debug level

mrogers950:
https://bugzilla.redhat.com/show_bug.cgi?id=1869293

### DIFF
--- a/cmd/manager/logcollector.go
+++ b/cmd/manager/logcollector.go
@@ -236,9 +236,9 @@ func reportOK(conf *daemonConfig, rt *daemonRuntime) {
 	}, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), maxRetries))
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/manager/logcollector.go
```diff
@@ -238,7 +238,7 @@ func reportOK(conf *daemonConfig, rt *daemonRuntime) {
 	if err != nil {
 		FATAL("Can't create configMap to report OK: '%v', aborting", err)
 	}
-	LOG("Created OK configMap '%s'", conf.LogCollectorConfigMapName)
+	DBG("Created OK configMap '%s'", conf.LogCollectorConfigMapName)
```
JAORMX, line 241: From a user that doesn't know the internals of the operator. Seeing a message "Created OK configMap" might be a little funky. Do you think we can rephrase this to something more explicit that would help folks debug this?

PR #99, opened 2020-08-26
Make default grace period to be 15 min

JAORMX:
The AIDE integrity checking is a very intensive process, with our
current default of 10s, the daemon would constantly take ~2.2G of memory
(for each instance... meaning, in each node).

This is excessive and will certainly raise alarms with folks running
this in production.

I propose to set our default grace period to be 900 (15 min), which
will result in the daemonsets taking 60MB when idle, and having usage
spikes every 15 mins.

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -43,6 +43,7 @@ const (
 	testConfDataKey        = "conf"
 	nodeWorkerRoleLabelKey = "node-role.kubernetes.io/worker"
 	mcWorkerRoleLabelKey   = "machineconfiguration.openshift.io/role"
+	defaultTestGracePeriod = 20
 )
 
 var mcLabelForWorkerRole = map[string]string{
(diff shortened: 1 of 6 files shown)


JAORMX:
Pull-request updated, HEAD is now b8e5370e3fc313346294e4f7c3a97f442df303bc

JAORMX:
Pull-request updated, HEAD is now 19e679aae9788f3a150c34db896789b47e7db3e2

JAORMX:
Pull-request updated, HEAD is now 53dfed67a8499aa01bbf86400626ad61d019e487

tests/e2e/helpers.go
```diff
@@ -405,7 +407,9 @@ func setupTest(t *testing.T) (*framework.Framework, *framework.Context, string)
 		},
 		Spec: fileintv1alpha1.FileIntegritySpec{
 			NodeSelector: nodeLabelForWorkerRole,
-			Config:       fileintv1alpha1.FileIntegrityConfig{},
+			Config: fileintv1alpha1.FileIntegrityConfig{
+				GracePeriod: 20,
```
mrogers950, line 411: Check out TestFileIntegrityChangeGracePeriod(), which has an initial check that DefaultGracePeriod has applied, so this change would make that test fail.

JAORMX:
Pull-request updated, HEAD is now 2010f0e9ac533c61b6e49f0b3ccb6810e11cc2fa

PR #104, opened 2020-09-15
Bug 1861303: Migrate holdoff logic to daemonset

JAORMX:
Having separate logic in the node controller to holdoff the integrity
checks, this patch migrates that functionality to the daemon itself.

This instead adds a watcher for the file integrity instance as part of
the daemon (having its own loop), and now the holdoff functionality
solely relies on the fileintegrity instance having the appropriate
annotations. So no need to check files anymore.


JAORMX:
Pull-request updated, HEAD is now d6489467132be9786ee60402add042205879d3f8

JAORMX:
Pull-request updated, HEAD is now 96233c96dee4abbecdd6c1684ac9333a1e2772ae

PR #105, opened 2020-09-15
Bug 1866805: Add notes to README

mrogers950:
Some more information about deletion of FI and manually re-initializing


JAORMX (commented):
/lgtm

PR #106, opened 2020-09-16
api: Add description to spec.config keys to enable oc explain

jhrozek:
Wording suggestions welcome, in general having the strings available for oc explain increases usability.


JAORMX (commented):
/lgtm

PR #108, opened 2020-09-21
Bug 1826301: Mark the nodeStatus and the FI object as Error in case the config can't even be loaded

jhrozek:
In case the configuration was wrong, the reinitLoop function would just
continue its loop, but because each loop tries to acquire the aideFiles
mutex, subsequent loops would just deadlock waiting on a mutex they
locked themselves.

To make sure the admin is notified about this, we watch for aide error conditions and put their string representation as a result to the fileIntegrityNodeStatus.

The fileIntegrityNodeStatus API is also extended with a lastResult attribute. This is the Result attribute with the latest timestamp. The reason I added the new attribute is that 1) it is easier this way to check if the "current" status is an error which is what the next patch does and 2) it will be useful in order to provide custom column printers

Finally, the status controller is extended so that if any of the fileIntegrityNodeStatus objects' lastResult is an Error, then the FI status as a whole is also an error so that the admin is notified that the FI does not work.

### DIFF
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -287,6 +287,7 @@ func reinitLoop(rt *daemonRuntime, conf *daemonConfig, exit chan bool) {
 			if err := runAideInitDBCmd(); err != nil {
 				LOG(err.Error())
 				time.Sleep(time.Second)
+				rt.UnlockAideFiles("reinitLoop")
 				continue
 			}
 			LOG("initialization finished")
@@ -329,6 +330,7 @@ func reinitLoop(rt *daemonRuntime, conf *daemonConfig, exit chan bool) {
 			if err := runAideInitDBCmd(); err != nil {
 				LOG(err.Error())
 				time.Sleep(time.Second)
+				rt.UnlockAideFiles("reinitLoop")
 				continue
 			}
 
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"sync"
-	"syscall"
 	"time"
 
 	"k8s.io/api/events/v1beta1"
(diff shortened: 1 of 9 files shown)


jhrozek:
this is related to the scenario described in https://bugzilla.redhat.com/show_bug.cgi?id=1826301

While the PR fixes the deadlock, I marked it as WIP because it doesn't surface the issue to the admin in a meaningful way. I can see several options:
 1. Just issue an event. Easy, but not super discoverable
 2. Issue an event and quit. Easy, very discoverable, but not very user-friendly.
 3. Handle the error codes that aide documents in `man aide` in the logcollector and create an error configmap. This is my preferred way. The only issue I have is that in this case the FI object still says active even if it's technically not.

I'd welcome other opinions.

JAORMX:
> this is related to the scenario described in https://bugzilla.redhat.com/show_bug.cgi?id=1826301
> 
> While the PR fixes the deadlock, I marked it as WIP because it doesn't surface the issue to the admin in a meaningful way. I can see several options:
> 
>     1. Just issue an event. Easy, but not super discoverable
> 
>     2. Issue an event and quit. Easy, very discoverable, but not very user-friendly.
> 
>     3. Handle the error codes that aide documents in `man aide` in the logcollector and create an error configmap. This is my preferred way. The only issue I have is that in this case the FI object still says active even if it's technically not.
> 
> 
> I'd welcome other opinions.

3. indeed sounds like the best option. Wouldn't there be a way for us to detect the configMap in the controller and update the FI object from there?

jhrozek:
New patches were pushed. They are still WIP, but this iteration at least surfaces the issue up to the `fileintegritynodestatus` object. Next up, updating the FI object if the nodestatus object contains an error condition.

jhrozek:
Now the FI object gets a new Error state which is used if any of the nodeStatus objects end up as error. Next up: e2e test and fix the config reloading otherwise the admin doesn't have an easy way out of the error condition.

jhrozek:
Now with tests. I think I'll do the config reloading separately, so I'm removing the WIP label.

JAORMX (approved):

PR #109, opened 2020-09-21
daemon: Shortcut the integrityInstanceLoop if a watch can't be found

jhrozek:
Especially in the case the watch command fails, the watcher variable
would have been nil and while we write to the exit channel to signal
that the daemon should quit, we would still continue in the
integrityInstanceLoop goroutine and potentially dereference the nil
value.

Let's cleanly exit from the goroutine instead.


JAORMX (approved):
/lgtm

PR #110, opened 2020-09-24
Remove replaces: from CSV

mrogers950:
This file is sourced for the downstream bundle builds now. Since the downstream bundle builds specify a single release it errors out on the replaces: line.

### DIFF
--- a/cmd/manager/logcollector.go
+++ b/cmd/manager/logcollector.go
@@ -284,7 +284,8 @@ func logCollectorMainLoop(rt *daemonRuntime, conf *daemonConfig, ch chan bool) {
 	for {
 		lastResult := <-rt.result
 		// We haven't received a result yet.
-		if lastResult == -1 {
+		// Or return code 18 - AIDE ran prior to there being an aide database.
+		if lastResult == -1 || lastResult == 18 {
 			DBG("No scan result available")
 			time.Sleep(time.Duration(conf.Interval) * time.Second)
 			continue
--- a/cmd/manager/logcollector.go
+++ b/cmd/manager/logcollector.go
@@ -118,9 +118,9 @@ func matchFileChangeRegex(contents []byte, regex string) string {
 }
 
 func annotateFileChangeSummary(contents []byte, annotations map[string]string) {
-	annotations[common.IntegrityLogFilesAddedAnnotation] = matchFileChangeRegex(contents, `\s+Added files:\s+(?P<num_added>\d+)`)
-	annotations[common.IntegrityLogFilesChangedAnnotation] = matchFileChangeRegex(contents, `\s+Changed files:\s+(?P<num_changed>\d+)`)
-	annotations[common.IntegrityLogFilesRemovedAnnotation] = matchFileChangeRegex(contents, `\s+Removed files:\s+(?P<num_removed>\d+)`)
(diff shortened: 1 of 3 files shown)


mrogers950:
Moving to centos:8 uncovered an issue that we would hit on the downstream builds, added a commit to fix that.

mrogers950:
Weird failure:
```
    helpers.go:582: dbg: wait for fail, result v1alpha1.FileIntegrityScanResult{LastProbeTime:v1.Time{Time:time.Time{wall:0x0, ext:63736570722, loc:(*time.Location)(0x2653f00)}}, Condition:"Failed", ResultConfigMapName:"aide-ds-test-check-ip-10-0-149-144.ec2.internal-failed", ResultConfigMapNamespace:"openshift-file-integrity", ErrorMsg:"", FilesAdded:0, FilesChanged:0, FilesRemoved:0}
    e2e_test.go:58: Expected one file to change, got 0
    e2e_test.go:66: configMap 'aide-ds-test-check-ip-10-0-149-144.ec2.internal-failed' does not have a failure log. Got: map[string]string{"integritylog":"Start timestamp: 2020-09-24 18:58:10 +0000 (AIDE 0.16)\nAIDE initialized database at /hostroot/etc/kubernetes/aide.db.gz\n\nNumber of entries:\t30480\n\n---------------------------------------------------\nThe attributes of the (uncompressed) database(s):\n---------------------------------------------------\n\n/hostroot/etc/kubernetes/aide.db.gz\n  MD5      : TgdV2mSz/zQ75W6+WTHNFg==\n  SHA1     : VuDEjskdA0qM0I4UdBKKkTN/K84=\n  RMD160   : WyauY49fTEPDwt9Y/q+QJeHQ2/U=\n  TIGER    : cpSLXD88rGlFhgwDRcL6x3kzLzreitVl\n  SHA256   : GV+YrgxHZMqpSJ5bxv2BV6Qz20/pwWKF\n             kXm4UQUPrXw=\n  SHA512   : 5jkDcyZ7allv/1W+ysfbdkmCH5EnQf7v\n             D0v2IfhRNTaNkDTMr+FfEBytPmqs5CDN\n             kWmxBHpNR4e3AApx5+cGNQ==\n\n\nEnd timestamp: 2020-09-24 18:58:42 +0000 (run time: 0m 32s)\n"}

```

mrogers950:
The last commit could indicate that the erroneous failure configMap might be from a race between initializing and the aide check running, if the tests pass

mrogers950:
This might need the fixes from https://github.com/openshift/file-integrity-operator/pull/108

cmd/manager/logcollector.go
```diff
@@ -284,7 +284,8 @@ func logCollectorMainLoop(rt *daemonRuntime, conf *daemonConfig, ch chan bool) {
 	for {
 		lastResult := <-rt.result
 		// We haven't received a result yet.
-		if lastResult == -1 {
+		// Or return code 18 - AIDE ran prior to there being an aide database.
+		if lastResult == -1 || lastResult == 18 {
```
jhrozek, line 288: Alternatively, what do you think about not sending error codes (anything return value larger than 4 if man 1 aide can be trusted) to rt.result, but instead handle the errors where they appear. We could also do exactly the other way around and always send the errors to the channel and handle them all in log collector, but I think handling the errors where they appear gives you more context around them, even if there is some code duplication.

jhrozek: but in general I agree that finishing 108 would be nice, I'll try to get the final version of the patches submitted as soon as possible.

PR #112, opened 2020-09-28
Bug 1869293: Add timestamp to daemon logs

mrogers950:
Also adding a commit to capture daemon logs in the E2E, same code as compliance-operator.

### DIFF
--- a/cmd/manager/logcollector.go
+++ b/cmd/manager/logcollector.go
@@ -51,7 +51,7 @@ const (
 var debugLog bool
 
 func LOG(format string, a ...interface{}) {
-	fmt.Printf(format+"\n", a...)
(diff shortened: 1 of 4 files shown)


cmd/manager/logcollector.go
```diff
@@ -51,7 +51,7 @@ const (
 var debugLog bool
 
 func LOG(format string, a ...interface{}) {
-	fmt.Printf(format+"\n", a...)
+	fmt.Printf(fmt.Sprintf("%s: %s\n", time.Now().Format(time.RFC3339), format), a...)
```
jhrozek, line 54: Why only add the timestamps to LOG and not DBG?

mrogers950: DBG uses LOG (see below)

PR #114, opened 2020-09-29
e2e: Be more resilient when setting node taints

JAORMX:
In cases where setting the node's taint fails, we should be more
resilient and try again.


jhrozek (approved):
/lgtm

PR #116, opened 2020-09-30
Reload FI when the configMap the FI points to changes

jhrozek:
The FI controller used to only watch for FI object changes. That meant
that if the configMap pointed to by spec.config changed, the operator
would not re-render the configuration the DS uses until the FI object
was 'touched'.

This patch uses a custom mapper to reconcile FI objects when CMs that
are used by a FI changes.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -434,12 +434,21 @@ func TestFileIntegrityBadConfig(t *testing.T) {
 			t.Fatal(err)
 		}
 	}()
+	defer logContainerOutput(t, f, namespace, testIntegrityName)
 
 	// Install the different config
 	createTestConfigMap(t, f, testIntegrityName, testConfName, namespace, testConfDataKey, brokenAideConfig)
 
 	// wait to go to error state.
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_test.go
```diff
@@ -443,6 +444,14 @@ func TestFileIntegrityBadConfig(t *testing.T) {
 	if err != nil {
 		t.Errorf("Timeout waiting for scan status")
 	}
+
+	// Fix the config
+	updateTestConfigMap(t, f, testConfName, namespace, testConfDataKey, testAideConfig)
+	// wait to go to error state.
```
mrogers950, line 450: s/error/active

tests/e2e/helpers.go
```diff
@@ -542,6 +542,20 @@ func createTestConfigMap(t *testing.T, f *framework.Framework, integrityName, co
 	updateFileIntegrityConfig(t, f, integrityName, configMapName, namespace, key, time.Second, 2*time.Minute)
 }
 
+func updateTestConfigMap(t *testing.T, f *framework.Framework, configMapName, namespace, key, data string) {
+	// create a test AIDE config configMap
```
mrogers950, line 546: we're getting/updating here not creating

PR #117, opened 2020-09-30
Several small cleanup patches

jhrozek:
They just help to control my OCD.

### DIFF
--- a/pkg/controller/node/node_controller.go
+++ b/pkg/controller/node/node_controller.go
@@ -24,11 +24,6 @@ import (
 
 var log = logf.Log.WithName("controller_node")
 
-const (
-	addHoldOffScriptCMName    = "addholdoff"
-	removeHoldOffScriptCMName = "rmholdoff"
-)
-
 // Add creates a new Node Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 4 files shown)


pkg/controller/node/node_controller.go
```diff
@@ -136,13 +131,13 @@ func (r *ReconcileNode) Reconcile(request reconcile.Request) (reconcile.Result,
 func (r *ReconcileNode) findRelevantFileIntegrities(currentnode *corev1.Node) []*fiv1alpha1.FileIntegrity {
 	resultingFIs := []*fiv1alpha1.FileIntegrity{}
 	fiList := fiv1alpha1.FileIntegrityList{}
-	r.client.List(context.TODO(), &fiList)
+	_ = r.client.List(context.TODO(), &fiList)
```
JAORMX, line 134: It's actually not entirely clear to me why we're ignoring errors here :confused:

mrogers950: Yeah we should handle them

jhrozek: I thought ignoring was acceptable since we'd end up with an empty list and at worst do nothing, but yeah, handling the errors is safer.

PR #124, opened 2020-12-14
Disable custom namespace in Makefile

JAORMX:
CI is using its own namespace and the Makefile isn't ready for it.


jhrozek (approved):
/lgtm

PR #125, opened 2021-01-04
Fix dev deployments in local clusters

JAORMX:
The dev workflow to deploy to a local cluster is broken. This fixes that.

* Non-kubeadmin users were not able to deploy
* We were not using a registry explicitly (podman needs it)

This also fixes the release process to use skipRange, similar to
what we did in the compliance-operator.


JAORMX:
Pull-request updated, HEAD is now 090de33785bed69f3b1cc50985a8f68b09e4eea2

PR #139, opened 2021-02-01
README.md: Add note about log compression

mrogers950:


README.md
```diff
@@ -147,6 +147,7 @@ Detailed information about changes:
 File: /hostroot/etc/resolv.conf
  SHA512   : Xl2pzxjmRPtW8bl6Kj49SkKOSBVJgsCI , tebxD8QZd/5/SqsVkExCwVqVO22zxmcq
 ```
+AIDE logs over 1mb are gzip compressed and base64 encoded. In this case, you will want to pipe the output of the above command to `base64 -d | gunzip`.
```
JAORMX, line 150: We should probably explain why this happens too.

JAORMX (changes requested):
lgtm, but we should explain a little more so folks don't think it was a random design decision.

README.md
mrogers950: done!

PR #149, opened 2021-02-10
daemon: Changes for running with pprof

mrogers950:
- Add local, pprof(heap), and aide config/data path options to the AIDE daemon to
  allow running standalone.
- Change default mount path of the aide config for the daemonSet to /config.
  This leaves /tmp available as an EmptyDir mount needed for pprof.
- Add golang to build/Dockerfile for running pprof in a Pod exec command.

### DIFF
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -18,10 +18,16 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
+	"path"
 	"sync"
 	"time"
+
+	"k8s.io/client-go/tools/clientcmd"
+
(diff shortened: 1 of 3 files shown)


cmd/manager/daemon.go
```diff
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
+	_ "net/http/pprof"
```
JAORMX, line 33: You'd need to make the daemonset a webserver to make this work. Currently it only listens for events from the kube api and does not expose a server you could query.

PR #159, opened 2021-03-04
Bug 1925105: Fix madvise setting

mrogers950:
The option is `GODEBUG=madvdontneed=1` not `GODEBUG=madvisedontneed=1`


jhrozek (approved):
/lgtm

PR #175, opened 2021-05-17
Exclude the CNI plugin directory

mrogers950:
This stops /etc/kubernetes/cni/net.d/ from being covered under scans with the
default config.


JAORMX (approved):
/lgtm

Issue #68, opened 2020-05-29
Set resource limits for containers


bigg01:
As a Platform Engineer I need to control usage of CPU Memory per Container. 
Please add ressource limits:

```
resources:
          requests:
            cpu: 100m
            memory: 100Mi
          limits:
            memory: 200Mi
            cpu: 400m
```

The Aide PODS were running for a day and used 1.6 Gb of memory for no reason.

cheers

jhrozek:
Thank you for filing the issue. We'll look into it next sprint.
While the resource limits are something we wanted to set either way, we also want to see if we can find the root cause of the leak.

felixkrohn:
Would it be possible to re-open this issue? After a week running the pods consume about 3GiB RAM each.
Current workaround could be to set namespaced defaults, but I find this less elegant.

JAORMX:
@feli

JAORMX:
@felixkrohn what versoin are you using?

felixkrohn:
0.1.13 as distributed by RH on operatorhub (image: http://quay.io/file-integrity-operator/file-integrity-operator:0.1.13)

felixkrohn:
Is there anything I can do to help you debug this? (we're not yet running it in production)
![image](https://user-images.githubusercontent.com/17177764/114354666-d4925600-9b6e-11eb-8985-f93d7accbaf2.png)

JAORMX:
@felixkrohn we'll look into it.

mrogers950:
@felixkrohn would you be able to run with the steps outlined in https://mrogers950.gitlab.io/openshift/2021/04/12/fio-profile/ ? 
It will enable pprof for the ds pods, but requires a container build from source. If you can capture the heap data at a few points (like once a few days in, then again the next week), that could be useful for us to take a look at. I've traced the same slow leak myself and it would be good to have a comparison.

felixkrohn:
@mrogers950 Thanks to the great how-to 👍 I got it running, and will send you the .gz files next week (don't hesitate to remind me should I forget...)

felixkrohn:
Did the traces help in any way?
Would it be OK to add memory limits (something between 500 and 1000M) to the f-i-o deployment, or do you expect this could cause unwanted side effects or even reduce reliability of the results?

mrogers950:
@felixkrohn yes, thanks for your help! the pprof data shows what I expected, which is that the daemon's actual heap usage is only a small percentage of the total that is reported by the cluster:
[...]

PR #176, opened 2021-05-18
Add operator and aide-ds pod limits

mrogers950:
Closes https://github.com/openshift/file-integrity-operator/issues/68

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -22,6 +22,13 @@ spec:
           args:
           - operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              memory: "10Mi"
+              cpu: "10m"
+            limits:
+              memory: "200Mi"
+              cpu: "100m"
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
--- a/pkg/controller/fileintegrity/fileintegrity_controller.go
+++ b/pkg/controller/fileintegrity/fileintegrity_controller.go
@@ -6,6 +6,8 @@ import (
 	"reflect"
 	"strconv"
 
+	"k8s.io/apimachinery/pkg/api/resource"
+
 	"github.com/go-logr/logr"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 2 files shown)


pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -673,6 +695,16 @@ func aideDaemonset(dsName string, fi *fileintegrityv1alpha1.FileIntegrity) *apps
 									MountPath: "/tmp",
 								},
 							},
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceMemory: resource.MustParse("100Mi"),
+									corev1.ResourceCPU:    resource.MustParse("100m"),
```
JAORMX, line 701: let's reduce the requests to:
```
 memory: 10Mi
 cpu: 10m
```

That'll make it more likely that the kubelet can schedule this. If it's too high sometimes it happens that the kubelet is unable to schedule the pod at all.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -673,6 +695,16 @@ func aideDaemonset(dsName string, fi *fileintegrityv1alpha1.FileIntegrity) *apps
...
 								},
 							},
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceMemory: resource.MustParse("10Mi"),
+									corev1.ResourceCPU:    resource.MustParse("10m"),
+								},
+								Limits: corev1.ResourceList{
+									corev1.ResourceMemory: resource.MustParse("300Mi"),
+									corev1.ResourceCPU:    resource.MustParse("400m"),
```
JAORMX, line 705: for the SNO case, I'd rather limit this too 100m. It might make integrity checks slower, but it allows us for better bin-packing. e.g. platform or control plane components get to take less space.

deploy/operator.yaml
```diff
@@ -22,6 +22,13 @@ spec:
           args:
           - operator
           imagePullPolicy: Always
+          resources:
+            requests:
+              memory: "20Mi"
```
JAORMX, line 27: The controllers are mostly sleeping. I suggest lowering this to 10Mi.

JAORMX (changes requested):

JAORMX (approved):
/lgtm

mrogers950:
The compressed log scan took quite a bit longer it seems
/retest

JAORMX:
If it takes too much of an impact we should increase the limit again then :/

PR #178, opened 2021-05-19
daemon: Handle SIGTERM and SIGKILL

mrogers950:
Just pushing what I came up with at the moment. Still need to test.

### DIFF
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -21,7 +21,9 @@ import (
 	"net/http"
 	"net/http/pprof"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/openshift/file-integrity-operator/pkg/common"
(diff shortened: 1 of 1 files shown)


cmd/manager/daemon.go
```diff
@@ -264,6 +266,8 @@ func daemonMainLoop(cmd *cobra.Command, args []string) {
 
 	errChan := make(chan error)
 	ctx, cancel := context.WithCancel(context.Background())
+	sigTermChan := make(chan os.Signal)
+	signal.Notify(sigTermChan, syscall.SIGTERM, syscall.SIGKILL)
```
JAORMX, line 270: Would [NotifyContext](https://golang.org/pkg/os/signal/#NotifyContext) be useful here?

mrogers950: It would indeed! I need to update deps to 1.16 first

PR #182, opened 2021-06-01
Bug 1966566: Handle AIDE error code 255

mrogers950:
Using a FIPS disallowed algorithm in the AIDE config under FIPS mode makes AIDE throw a 255 error code. Handle this error as a possible FIPS error, even though it could probably be encountered in some other rare case, but this would be the most informative to our user.

### DIFF
--- a/pkg/common/util.go
+++ b/pkg/common/util.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"strconv"
 
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -22,6 +24,8 @@ type FileIntegrityComponent uint
 const (
 	OPERATOR = iota
 )
+
+const AIDE_RETFAIL = 255
 
 var componentDefaults = []struct {
 	defaultImage string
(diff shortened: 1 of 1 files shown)


mrogers950:
Updated to include checking if FIPS mode is enabled to adjust the error message accordingly.

pkg/common/util.go
```diff
@@ -188,8 +190,50 @@ func getExitCode(runCmdErr error, defaultError int) int {
...
+
+	d, err := ioutil.ReadAll(f)
+	if err != nil {
+		if closeErr := f.Close(); closeErr != nil {
+			return false, closeErr
+		}
+		return false, err
+	}
+
+	fipsStatus, err := strconv.Atoi(string(d[0]))
```
jhrozek, line 207: Do you think we should protect against `d` being empty?

mrogers950: fixed

pkg/common/util.go
```diff
@@ -216,12 +272,17 @@ var aideErrLookup = []struct {
...
 
 func GetAideErrorMessage(rv int) string {
 	if rv < aideErrBase || rv > aideErrSentinel {
-		// default to the sentinel error message for unknown or unexpected errors
-		rv = aideErrSentinel
+		// default to the sentinel error message for unknown or unexpected errors (two exceptions that are outside the
+		// expected range are the FIPS errors)
+		if rv != aideErrFips && rv != aideErrPossibleFips {
+			rv = aideErrSentinel
+		}
```
jhrozek, line 285: So I haven't tried this patch, but if the numerical value of the fips errors are 254, then the following line does `rv -= aideErrBase` then the index will be outside the `aideErrLookup` array, right?

(looking at how we handle the error code to message translation, I wrote this and this was clearly me still thinking in C but somehow writing Go..we might better use a proper data structure instead of magic numbers that somehow happen to point to a string array...)

jhrozek (commented):
just this nit and then it'll look good

pkg/common/util.go
mrogers950: @jhrozek See the new update, I now cover the aideErrFips and aideErrPossibleFips errors in the range of the sentinel value., and GetAideExitCode() converts the 255 error from AIDE into one of those two (instead of trying to shoe it into the lookup array.).

JAORMX (approved):
/approve

Waiting for @xiaojiey to check this out before merging.

PR #184, opened 2021-06-22
Add controller prometheus metrics

mrogers950:


deploy/role.yaml
```diff
@@ -34,6 +34,13 @@ rules:
   verbs:
   - get
   - create
+  - update
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - prometheusrules
+  verbs:
+  - create
```
JAORMX, line 43: why do we need these? I don't see in the source code where the operator would create a PrometheusRule

mrogers950: That's at https://github.com/openshift/file-integrity-operator/pull/184/files#diff-51bd220db414457ea4470ea5d71465bd07e1c789d156c0104e1fce2f3ddbcb1aR224

JAORMX (approved):
/lgtm

yeesss!!!

PR #185, opened 2021-07-09
Enable TLS for controller metrics

mrogers950:
Plus a bit of refactoring


deploy/service_account.yaml
```diff
@@ -42,3 +42,30 @@ metadata:
   annotations:
     kubernetes.io/service-account.name: default
 type: kubernetes.io/service-account-token
+---
+apiVersion: v1
+kind: Service
```
JAORMX, line 47: Are we able to ship this via the OLM?

deploy/operator.yaml
```diff
@@ -42,6 +42,14 @@ spec:
...
             - name: RELATED_IMAGE_OPERATOR
               value: "quay.io/file-integrity-operator/file-integrity-operator:latest"
+          volumeMounts:
+            - name: file-integrity-operator-metrics
+              mountPath: /etc/tls/pki
+              readOnly: true
+      volumes:
+        - name: file-integrity-operator-metrics
+          secret:
+            secretName: file-integrity-operator-metrics
```
JAORMX, line 52: Where are we getting this secret from? Are we able to auto-generate it from the OLM?

mrogers950: From the service-ca (service.beta.openshift.io/serving-cert-secret-name annotation on the service).

I've got some further changes to the PR in flight.  I should also be able to use the namespace annotation for monitoring instead of the user-namespace method, like you mentioned. I'm missing the role needed for the prometheus permissions,

JAORMX:
/lgtm

nice!

PR #187, opened 2021-08-02
Move test-specific permissions out of deploy manifests

mrogers950:
We only need the rules and stuff for fio-metrics-client while testing, and not normal deploys.


mrogers950:
@Vincent056

Vincent056 (commented):
/lgtm

PR #189, opened 2021-08-05
Use fedora-minimal:34 base image

mrogers950:
The centos:8 one previously used was quite hefty, switch to fedora-minimal.
This uses :34 instead of :latest to avoid moving to the next version by
surprise.


mrogers950:
Added a commit to update the CI image too.

PR #190, opened 2021-08-11
Bug 1862022: Per-node reinit during update

mrogers950:
Make re-init daemonSets apply to a single node at a time when going through a
node update. This fixes the issue where all nodes are re-initialized each time
a single node updates.

### DIFF
--- a/pkg/controller/fileintegrity/fileintegrity_controller.go
+++ b/pkg/controller/fileintegrity/fileintegrity_controller.go
@@ -2,6 +2,7 @@ package fileintegrity
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -150,9 +151,33 @@ func (r *ReconcileFileIntegrity) handleDefaultConfigMaps(logger logr.Logger, f *
 	return confCm, scriptsUpdated, nil
 }
 
-func (r *ReconcileFileIntegrity) createReinitDaemonSet(instance *fileintegrityv1alpha1.FileIntegrity) error {
+// narrowNodeSelector returns a LabelSelector as-is from instance if node == nil, otherwise, return a LabelSelector for
+// node if node is contained in the node group. Used for making a node-specific DS.
+func (r *ReconcileFileIntegrity) narrowNodeSelector(instance *fileintegrityv1alpha1.FileIntegrity, node *corev1.Node) (metav1.LabelSelector, error) {
+	narrowSelector := metav1.LabelSelector{}
+	if node == nil {
+		narrowSelector.MatchLabels = instance.Spec.NodeSelector
+		return narrowSelector, nil
+	}
+
+	hostnameFromLabel, ok := node.Labels["kubernetes.io/hostname"]
+	if !ok || len(hostnameFromLabel) == 0 {
+		return narrowSelector, errors.New("couldn't find hostname label from node")
+	}
+
+	narrowSelector.MatchLabels = map[string]string{
+		"kubernetes.io/hostname": hostnameFromLabel,
+	}
+
+	return narrowSelector, nil
+}
+
+// createReinitDaemonSet creates a re-init daemonSet. The daemonSet will be a single-node reinit if node is not "" (like
+// in the node update case), otherwise, it will apply to the whole node instance node group (like in the manual re-init
+// case).
+func (r *ReconcileFileIntegrity) createReinitDaemonSet(instance *fileintegrityv1alpha1.FileIntegrity, node string) error {
 	daemonSet := &appsv1.DaemonSet{}
-	dsName := common.GetReinitDaemonSetName(instance.Name)
+	dsName := common.ReinitDaemonSetNodeName(instance.Name, node)
 	dsNamespace := common.FileIntegrityNamespace
 
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: dsName, Namespace: dsNamespace}, daemonSet)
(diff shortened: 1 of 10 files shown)


mrogers950:
@jhrozek @JAORMX please review when you can!

pkg/controller/status/status_controller.go
```diff
@@ -27,7 +28,7 @@ import (
 )
 
 var log = logf.Log.WithName("controller_status")
-var statusRequeue = time.Second * 30
+var statusRequeue = time.Second * 10
```
JAORMX, line 31: why the 10 second change?

JAORMX (commented):
overall this lgtm, just got one question, but I'll give this another review anyway.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -324,14 +361,15 @@ func (r *ReconcileFileIntegrity) Reconcile(request reconcile.Request) (reconcile
 		return reconcile.Result{}, err
 	}
 
-	_, forceReinit := instance.Annotations[common.AideDatabaseReinitAnnotationKey]
+	nodeToReinit := ""
+	nodeToReinit, forceReinit := instance.Annotations[common.AideDatabaseReinitAnnotationKey]
```
JAORMX, line 365: let's handle a case where `nodeToReinit` is an empty string (as it used to be). This way, in case there's a FIO upgrade in the middle of a re-init, we won't get this funky case. This could also happen if a user (for some weird reason) changes the value of the label to empty string.

pkg/controller/status/status_controller.go
mrogers950: TBH I forgot I changed this while debugging. The important piece to this was that I added the re-queue on successful status update back. When I removed it some patches ago, I thought it was affecting things differently.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -150,9 +151,39 @@ func (r *ReconcileFileIntegrity) handleDefaultConfigMaps(logger logr.Logger, f *
...
+// node if node is contained in the node group. Used for making a node-specific DS.
+func (r *ReconcileFileIntegrity) narrowNodeSelector(instance *fileintegrityv1alpha1.FileIntegrity, node string) (metav1.LabelSelector, error) {
+	narrowSelector := metav1.LabelSelector{}
+	if len(node) == 0 {
+		narrowSelector.MatchLabels = instance.Spec.NodeSelector
+		return narrowSelector, nil
+	}
+
+	n := &corev1.Node{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: node}, n)
```
JAORMX, line 164: let's handle if this call returns NotFound. This way we can dismiss a Node that has been scaled down and no longer exists.

JAORMX (changes requested):

pkg/controller/fileintegrity/fileintegrity_controller.go
mrogers950: that's handled inside createReinitDaemonSet(), the result of an empty string for node there is that you get a re-init ds that applies to all FI nodes, as before.

pkg/controller/fileintegrity/fileintegrity_controller.go
mrogers950: Done.

pkg/controller/status/status_controller.go
```diff
@@ -121,26 +122,55 @@ func (r *ReconcileFileIntegrityStatus) Reconcile(request reconcile.Request) (rec
...
+	if getErr != nil {
+		return reconcile.Result{}, getErr
+	}
+
+	// If any of the associated nodes have an active re-init daemonSet, then the FI is initializing
+	for _, node := range nodeList.Items {
+		nodeName := node.Name
+		ds := &appsv1.DaemonSet{}
+		dsName := common.ReinitDaemonSetNodeName(instance.Name, nodeName)
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: dsName, Namespace: common.FileIntegrityNamespace}, ds)
```
JAORMX, line 136: don't we have relevant labels we could use to filter these? In that case you could just do a `List` call and not attempt to do a call per node, which would be very costly in a very big cluster.

pkg/controller/fileintegrity/fileintegrity_controller.go
JAORMX: Got it!

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -165,7 +190,27 @@ func (r *ReconcileFileIntegrity) createReinitDaemonSet(instance *fileintegrityv1
...
+		getErr := r.client.Get(context.TODO(), types.NamespacedName{Name: node}, &n)
+		if getErr != nil && !kerr.IsNotFound(getErr) {
+			return getErr
+		}
+		// Something happened to the node we wanted, continue without spawning the re-init daemonSet.
+		if kerr.IsNotFound(getErr) {
+			return nil
+		}
+		reinitNode = &n
+	}
```
JAORMX, line 205: Moving this block to its own function or commenting this a little more would help the reader know that having an empty nodeName is fine and an expected behavior.

JAORMX (changes requested):

jhrozek (approved):
I think the PR looks good as-is, only one question:
is the per-node re-init something you wanted to have documented as an external usage or only for the internal use when a node is initializing? If the former, then please update README.md and ask the docs writers to document this. But I'm not sure it is needed honestly.

JAORMX:
@jhrozek  I think it's an implementation detail. What users should care about is the reinitialization annotation, not the mechanism that goes behind it. But... maybe having developer docs would help us in the future maintenance-wise.

PR #192, opened 2021-08-19
Optimize per-node reinit calls

mrogers950:
- Use an owner label to look up relevant re-init daemonSets instead of
  individually by name. This simplifies the status controller's reconcile loop.

@JAORMX @jhrozek


mrogers950:
/hold
holding for acks

JAORMX (approved):
nice!

PR #195, opened 2021-08-30
Bug 1999186: Add MCO and CVO related config excludes

mrogers950:
@JAORMX

### DIFF
--- a/pkg/controller/fileintegrity/config_defaults.go
+++ b/pkg/controller/fileintegrity/config_defaults.go
@@ -27,6 +27,7 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 /hostroot/boot/        CONTENT_EX
 /hostroot/root/\..* PERMS
 /hostroot/root/   CONTENT_EX
+!/hostroot/root/\.kube
 !/hostroot/usr/src/
 !/hostroot/usr/tmp/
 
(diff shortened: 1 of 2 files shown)


pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -45,6 +45,9 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/selinux/targeted
 !/hostroot/etc/openvswitch/conf.db
 !/hostroot/etc/kubernetes/cni/net.d/*
+!/hostroot/etc/machine-config-daemon
+!/hostroot/etc/pki/ca-trust/extracted/java
+!/hostroot/etc/cvo
```
JAORMX, line 50: I don't think we should fully exclude these, but at least we should be checking that the appropriate permissions are kept.

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -45,6 +45,9 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/selinux/targeted
 !/hostroot/etc/openvswitch/conf.db
 !/hostroot/etc/kubernetes/cni/net.d/*
+!/hostroot/etc/machine-config-daemon
+!/hostroot/etc/pki/ca-trust/extracted/java
```
JAORMX, line 49: isn't this merely a PKCS12 file? that one shouldn't change as often.

JAORMX (changes requested):

pkg/controller/fileintegrity/config_defaults.go
mrogers950: Actually I think there's a regex pattern I can use to still watch for weird stuff in the directory and omit the dynamic files. Gotta look into it some more.

mrogers950:
@JAORMX I've updated it to a more specific exclude, so we should have better coverage with this iteration.

PR #196, opened 2021-09-01
minor: adapt prometheusrule to only alert on currently existing nodes

felixkrohn:
background: on a cluster with frequently changing nodes there's a lot of alert noise generated because of lingering fileintegritynodestatuses belonging to nodes that don't exist anymore.
Maybe a better approach would be to delete the fileintegritynodestatus here: https://github.com/openshift/file-integrity-operator/blob/a61147841dccfb5719f042476f144270a0d2551d/pkg/controller/node/node_controller.go#L101 (unfortunately that exceeds my read-only Golang capabilities)

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ spec:
         description: Node {{ $labels.node }} has an an integrity check status of Failed for
           more than 1 second.
         summary: Node {{ $labels.node }} has a file integrity failure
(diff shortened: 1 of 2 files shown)


mrogers950:
@felixkrohn thank you! good catch. 
We probably should delete lingering node statuses, although that would be done in a different controller. Let's merge this and I can do a follow up.

PR #198, opened 2021-09-13
Makefile: Rename IMAGE_FORMAT var

mrogers950:
IMAGE_FORMAT will be deprecated, so rename this to a variable we'll define in openshift/release


JAORMX (approved):
/lgtm

PR #203, opened 2021-10-05
Bug 2010706: Fix nil deref in daemonSet upgrade path

mrogers950:
@JAORMX @jhrozek


jhrozek (approved):
/lgtm

oops :-)
I also did a cursory git grep in both CO and FIO and it appears we don't have other occurences of this issue anywhere else.

PR #204, opened 2021-10-05
Makefile: Make image bundle depend on $TAG

mrogers950:
This ensures that pushing the bundle uses the replaced $TAG instead of 'latest'.


jhrozek (approved):
/lgtm

PR #205, opened 2021-10-05
Release v0.1.20

mrogers950:
### Changes
- Makefile: Make image bundle depend on $TAG
- Fix nil deref in daemonSet upgrade path


mrogers950:
@JAORMX @jhrozek

PR #207, opened 2021-10-21
Handle IO error in init database more leniently

JAORMX:
This retries after a default sleep if there's an IO error coming from
the init database function.

Signed-off-by: Juan Antonio Osorio Robles <jaosorior@redhat.com>


JAORMX:
Pull-request updated, HEAD is now be050f4021833aeb0489b6fdcc37e7eb10d73e6c

Issue #211, opened 2021-12-09
Move aide.reinit out of `/etc`


cgwalters:
This is related to  https://github.com/ostreedev/ostree/pull/2453 - see also https://bugzilla.redhat.com/show_bug.cgi?id=1945274

Basically it looks like this operator is writing a file into `/etc` at runtime.  Since this is for *dynamic* state, not persistent state, it should go in `/run`.  This would avoid a seen-in-the-wild race condition at shutdown time where ostree wants to propagate the modified config file, but the daemonset has deleted it in the meantime.

(Arguably, ostree should be OK with files vanishing in `/etc` at shutdown time, but OTOH it can mean it's capturing an inconsistent snapshot of `/etc)

cgwalters:
```
ostree[658445]: error: During /etc merge: Reading modified config file: fstatat(kubernetes/aide.reinit): No such file or directory
Nov 26 21:55:29 
```

jhrozek:
I believe this was fixed in 123722228e927e944419d1df0cb78f077869149e and even linked with that PR, not sure why is the issue still open.

PR #212, opened 2021-12-13
Bug 2033311: Move aide.reinit to /run

mrogers950:
Related to https://github.com/ostreedev/ostree/pull/2453, this moves the
aide.reinit file to /run (under hostroot) to avoid a possible race condition on
node shutdown.

https://github.com/openshift/file-integrity-operator/issues/211

@jhrozek @JAORMX

### DIFF
--- a/cmd/manager/daemon.go
+++ b/cmd/manager/daemon.go
@@ -37,8 +37,9 @@ import (
 )
 
 const (
-	defaultAideFileDir   = "/hostroot/etc/kubernetes"
-	defaultAideConfigDir = "/tmp"
+	defaultAideFileDir    = "/hostroot/etc/kubernetes"
+	defaultAideConfigDir  = "/tmp"
+	defaultAideRuntimeDir = "/hostroot/run"
 	// Files for checks and reading. We copy the AIDE output files to here.
 	aideReadDBFileName  = "aide.db.gz"
 	aideReadLogFileName = "aide.log"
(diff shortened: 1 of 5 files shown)


mrogers950:
> The PR looks clean enough but what about upgrades? Shouldn't we clean up the old files? or, if we don't won't the issue just keep hapenning on the affected clusters?

Good suggestion. Initially I thought it wouldn't make a difference. I will add the cleanup and some tests.

jhrozek:
/retest
it looks like the CI namespaces weren't created:
```
e2e_test.go:183: Verifying that /hostroot/etc/kubernetes/aide.reinit no longer exists
    helpers.go:1353: error getting output exit status 1
    e2e_test.go:185: Error from server (NotFound): namespaces "ci-op-7z5bihm2" not found
        
    e2e_test.go:187: /hostroot/etc/kubernetes/aide.reinit still exists
```

jhrozek:
Hmm, no, still the same error, so the test is buggy. Is the invocation to oc debug correct, shouldn't you call "oc debug sh -c ...shell command + arguments" instead of just "oc debug -- shell command + arguments"?
(Don't have a cluster at the moment, so just guessing)

mrogers950:
@jhrozek trying a different way to manipulate and test for the old reinit file. there's something about `oc debug node` that exec is not able to handle correctly, at least by default.

mrogers950:
@pdhamdhe @xiaojiey PTAL

Issue #201, opened 2021-09-21
Operator not cleaning up old aide.log.backup and aide.db.gz.backup files


ribua:
Presently we are running the file integrity operator , however it is filling up the /etc/kubernetes directory with files dating back from more than 3-4 months. Is there a way for us to clean up the files automatically after 30 or a certain  period so that it doesn't use up the space and fill up the disk drive.  The clean up using logrotate would also be a feasible option since it is already available in coreos. 

`sh-4.4# ls -al aide.db.gz.backup* |wc -l
252`

`sh-4.4# ls -al aide.log.backup* |wc -l
252`

mrogers950:
Thanks for the report, we definitely should have the operator clean up old entries.

alexisph:
We're using v0.1.16 and we have this issue as well.

mrogers950:
v0.1.23 has the `maxBackups` pruning option now.

PR #217, opened 2022-02-21
WIP - Prune old DB and log backup files

mrogers950:
Make the AIDE daemon retain only 5 sets of re-init backup files by deleting the
older ones.

Ref: https://github.com/openshift/file-integrity-operator/issues/201
Ref: https://issues.redhat.com/browse/CMP-1215


mrogers950:
@jhrozek PTAL

cmd/manager/daemon_util.go
```diff
@@ -254,6 +276,50 @@ func backupFile(file string) error {
...
+		return backups[i].ModTime().Unix() < backups[j].ModTime().Unix()
+	})
+
+	// Delete the older entries
+	last := len(backups) - max
+	for i, _ := range backups {
+		if i < last {
+			removeErr := os.Remove(path.Join(dirPath, backups[i].Name()))
+			if removeErr != nil {
+				return removeErr
```
jhrozek, line 314: Do you want to abort or just log the error and continue?

cmd/manager/daemon_util.go
```diff
@@ -254,6 +276,50 @@ func backupFile(file string) error {
...
+	}
+
+	if len(backups) <= max {
+		// no pruning needed
+		return nil
+	}
+
+	// Sort backups oldest first
+	sort.Slice(backups, func(i, j int) bool {
+		return backups[i].ModTime().Unix() < backups[j].ModTime().Unix()
```
jhrozek, line 305: You could also use ModTime().After/Before, but this works, too

cmd/manager/daemon_util.go
jhrozek: Or potentially save an error to a variable and return it only after trying to remove all backups

tests/e2e/helpers.go
```diff
@@ -1179,6 +1179,21 @@ func confirmRemovedLegacyReinitFile(t *testing.T, ctx *framework.Context, f *fra
 	return waitForPod(containerCompleted(t, f.KubeClient, pod.Name, namespace, 1))
 }
 
+// tests for a set of re-init backups when MaxBackups is set to 1.
+func confirmMaxBackupFiles(t *testing.T, ctx *framework.Context, f *framework.Framework, node, namespace string) error {
+	pod := privPodOnNode(namespace, "test-ls-pod", node, "test \"$( ls /etc/kubernetes/*backup-* | wc -l )\" -eq 2")
```
jhrozek, line 1184: Hmm wouldn't `ls` print multiple files per line by default? Note that there's also `ls -1` to print one file per line

build/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Step one: build file-integrity-operator
-FROM registry.access.redhat.com/ubi8/go-toolset as builder
+FROM golang:1.16 as builder
```
jhrozek, line 2: I'm fine with this, but why not just go to 1.17? Of course it's OK to do that in another patch if you prefer to make incremental bumps

jhrozek (commented):
Just some nits and questions, but overall looking good!

tests/e2e/helpers.go
jhrozek: Ugh the formatting makes the switch unredable but it's "dash-one".

build/Dockerfile
mrogers950: Yep, I'll update to 1.17 after, with CI image update too -  CI is still on 1.16

cmd/manager/daemon_util.go
mrogers950: Unix rules :)

cmd/manager/daemon_util.go
mrogers950: Yeah, It's probably better to not error out here, and log about it.

mrogers950:
@jhrozek should be all set now.

jhrozek (approved):
/lgtm
sorry the review took so long!

PR #218, opened 2022-03-11
Bug 2049206: Re-create AIDE configMaps with missing owner

mrogers950:
When upgrading from much older versions, the AIDE configMaps that correspond to
FileIntegrity instances were left without an owner label, and this broke
re-inits. Fix this by re-creating the configMaps if they are found without the
owner label.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -483,6 +483,34 @@ func TestFileIntegrityConfigurationIgnoreMissing(t *testing.T) {
 	}
 }
 
+// Ensure that the owner-label remains on the configmap.
+func TestFileIntegrityConfigMapOwnerUpdate(t *testing.T) {
+	f, testctx, namespace := setupTest(t)
+	testName := testIntegrityNamePrefix + "-ownerupdate"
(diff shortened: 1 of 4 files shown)


tests/e2e/e2e_test.go
```diff
@@ -483,6 +483,34 @@ func TestFileIntegrityConfigurationIgnoreMissing(t *testing.T) {
 	}
 }
 
+// Ensure that the owner-label remains on the configmap.
+func TestFileIntegrityConfigMapOwnerUpdate(t *testing.T) {
+	f, testctx, namespace := setupTest(t)
+	testName := testIntegrityNamePrefix + "-ownerupdate"
```
rhmdnd, line 489: Interesting - I thought this was just the name for the label, but it's actually the test name and it's consistent across tests. I was going to recommend naming this variable to something like `labelName` but that doesn't make sense.

The testing framework doesn't wire up unique test names?

rhmdnd (approved):
LGTM - I have one question on test names, but it's not specific to this patch.

Thanks, Matt!

jhrozek (approved):
/lgtm

PR #219, opened 2022-03-22
e2e: Add the operator logs to artifacts

mrogers950:
This allows us to fail individual e2e tests and still retain the operator logs
for each. Beforehand, we always skipped cleanup on error in order to get the
failing operator logs.


jhrozek:
/retest
cluster didn't install

PR #220, opened 2022-03-24
Break apart the make target for release

rhmdnd:
Previously, we used a single makefile target for doing a release (e.g.,
`make release`). This works well, but we started using a multi-step
process in the compliance-operator [0] that prepares the release and
gives the person proposing the release an opportunity to review the
release details before pushing them to the repository. This helps catch
issues before we change details of the upstream repository that could
impact how the code is consumed.

[0] https://github.com/openshift/compliance-operator/blob/master/doc/contributor.md#proposing-releases


mrogers950 (approved):
/lgtm

PR #224, opened 2022-04-07
Bug 2072058: Use correct data keys for script configMaps

mrogers950:
Previously when enforcing the default configMap script contents, we compared
the wrong data keys. This led to the aide-reinit script not being updated
properly on an operator upgrade.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=2072058


jhrozek (approved):
/lgtm

PR #231, opened 2022-05-24
Added GithubCI to build  multi-arch image

gauravpbankar:
Raising a PR for building multi-arch images through Github action CI.

Created Github CI which can have github action and qemu statis user on top of x86 arch and we are using binaries of Power and other archs which are enabling us to create Virtual env for those architectures by this way we are building multi-arch images.

Note: To make it work need to create below two secrets in github repo 

**- QUAY_USER ---> quay username**
**- QUAY_PWD  ---> quay repo password**

### DIFF
--- a/.github/workflows/multi-arch_images_build.yml
+++ b/.github/workflows/multi-arch_images_build.yml
@@ -1,0 +1,83 @@
+name: Multiarch build
+
+on:
+  push:
+      
+env:
+  IMAGE_NAME: file-integrity-operator
+  IMAGE_TAG: multi-arch
+  IMAGE_REGISTRY: quay.io
+  IMAGE_NAMESPACE: openshift-file-integrity
+
+jobs:
+  build:
+    name: Build and push multi-arch image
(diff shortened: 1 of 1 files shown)


mrogers950:
Hi @gauravpbankar, thanks for the PR. Unfortunately, the openshift github org does not have actions enabled. We've explored this before, and the response from the DPP team is that actions will not be enabled any time soon :(

gauravpbankar:
@mrogers950  if we are not going to support any CI for buiding multi-arch image then what will be the best approach to get support for Power?

JAORMX:
@gauravpbankar note that this would only be for upstream builds. Downstream builds are handled by an entirely different pipeline, so supporting Power is not limited by what's available in this GH repo/org.

mrogers950:
@gauravpbankar like @JAORMX mentioned, we can aim for supporting Power with images that we build internally for the Red Hat catalog. The quay.io images that we provide here are mostly for convenience and testing.

PR #234, opened 2022-05-26
Allow running under OLM's AllNamespaces install mode

mrogers950:
This is related to the "de-scoping" changes that OLM will see. It means that eventually OLM will only install operators using the AllNamespaces installModeType, which sets WATCH_NAMESPACE to empty, and assumes that the operator will watch for objects in all namespaces. This PR enables AllNamespaces as a supported type, and forces to Own/SingleNamespace mode when running. 

Note, forcing the OwnNamespace mode might seem like a hack, but we can't safely fully de-scope (switch to watching all namespaces) until OLM is able to restrict an operator to a single namespace through its new API (tentatively called "FeatureBinding"). Also, fully-descoping is only critical for operators that have dependencies, or depended on by operators with multiple installModes, or singleton operators that are multi-tenant aware.

Ref: https://issues.redhat.com/browse/OLM-1579
Ref: https://issues.redhat.com/browse/CMP-1355

### DIFF
--- a/pkg/common/var.go
+++ b/pkg/common/var.go
@@ -29,7 +29,9 @@ func GetOperatorNamespace() (string, error) {
 	return ns, nil
 }
 
-// GetWatchNamespace returns the Namespace the operator should be watching for changes
+// GetWatchNamespace returns the Namespace the operator should be watching for changes. Eventually the watch namespace
+// will not be used when OLM begins to support only the AllNamespaces install type. To support AllNamespaces initially,
+// GetWatchNamespace will return the operator namespace if WATCH_NAMESPACE is empty.
 func GetWatchNamespace() (string, error) {
 	// WatchNamespaceEnvVar is the constant for env variable WATCH_NAMESPACE
 	// which specifies the Namespace to watch.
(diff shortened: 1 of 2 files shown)


pkg/common/var.go
```diff
@@ -40,6 +42,14 @@ func GetWatchNamespace() (string, error) {
...
 	}
+
+	if ns == "" {
+		opns, err := GetOperatorNamespace()
+		if err != nil {
+			return "", err
+		}
+		ns = opns
+	}
 	return ns, nil
```
rhmdnd, line 53: Before - `ns` could have been an empty string, but now that's not possible unless we return an error on line 49.

If we never return an empty string under normal circumstances, will that restrict the operator behavior to only the operator namespace, or is it ignored now?

I'm trying to think through this from the perspective of a user who is using the operator to monitor all namespaces.

mrogers950: Right, this forces the watch namespace to always be the operator namespace. Since the operator only needs to watch for configMaps in its namespace for its own accounting, and the CRDs are only applicable cluster-wide, there's no need to watch for all or multiple namespaces.

pkg/common/var.go
```diff
@@ -29,7 +29,9 @@ func GetOperatorNamespace() (string, error) {
...
 }
 
-// GetWatchNamespace returns the Namespace the operator should be watching for changes
+// GetWatchNamespace returns the Namespace the operator should be watching for changes. Eventually the watch namespace
+// will not be used when OLM begins to support only the AllNamespaces install type. To support AllNamespaces initially,
+// GetWatchNamespace will return the operator namespace if WATCH_NAMESPACE is empty.
```
rhmdnd, line 34: Since OLM is only going to support AllNamespace models in the future, this will fail in the future if it were reverted, right?

I'm wondering if there is something we can do to cover that case in the meantime.

PR #238, opened 2022-06-27
tests: Expose -skipCleanupOnError as an env variable

jhrozek:
By default, our tests don't clean up after themselves which is a good
default. However, when iterating on tests, it might be irritating to
have to clean up manually each time, so let's expose that option as an
environment variable.

Jira: [CMP-1420](https://issues.redhat.com//browse/CMP-1420)


JAORMX (approved):

mp-redhat:
As suggested by @jhrozek ,  changing the -skipCleanupOnError=false in Makefile started working the continuous e2e run test.

jhrozek:
@Vincent056 @rhmdnd PTAL

rhmdnd (approved):

PR #240, opened 2022-06-27
bug 2101393: operator: Set namespace label for alert

jhrozek:
According to style guide
https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#style-guide
an alert should set namespace label to identify which component is raising
the alert however NodeHasIntegrityFailure is not setting namespace label:
https://github.com/openshift/file-integrity-operator/blob/master/cmd/manager/operator.go#L266

rhbz: #2101393


jhrozek:
/hold
for QE review

jhrozek:
> 

Looks like RHBZ doesn't have that TargetVersion yet. I moved the bug to POST manually.

PR #241, opened 2022-06-28
Update the change log to include changes for version 0.1.25

rhmdnd:


rhmdnd:
Actually - the `make prepare-release` target still handles updating the change log. I'm going to roll these changes into the commit that actually proposes the release.

PR #244, opened 2022-06-29
make: Add a deploy-local target

jhrozek:
Adds a deploy-local target that builds an image from the current tree,
pushes it into the cluster and deploys FIO.

To tear down FIO, use the "make tear-down" command.

Jira: CMP-1401


jhrozek:
@xiaojiey if you add the LGTM label, I can merge the PR :-)

PR #255, opened 2022-07-05
e2e: Use pod instead of ds for logcompress file toucher

mrogers950:
Also only modify and check on a single node, to help speed things up.

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -346,9 +346,14 @@ func setupTest(t *testing.T) (*framework.Framework, *framework.Context, string) 
 	return framework.Global, testctx, namespace
 }
 
-// setupFileIntegrity creates the FileIntegrity instance and confirms
+// setupFileIntegrity creates the FileIntegrity instance with the default grace period.
 func setupFileIntegrity(t *testing.T, f *framework.Framework, testCtx *framework.Context, integrityName, namespace string) {
-	testIntegrityCheck := newTestFileIntegrity(integrityName, namespace, nodeLabelForWorkerRole, defaultTestGracePeriod, true)
+	setupFileIntegrityWithGracePeriod(t, f, testCtx, integrityName, namespace, defaultTestGracePeriod)
+}
+
+// setupFileIntegrityWithGracePeriod creates the FileIntegrity instance and confirms that it goes active.
+func setupFileIntegrityWithGracePeriod(t *testing.T, f *framework.Framework, testCtx *framework.Context, integrityName, namespace string, grace int) {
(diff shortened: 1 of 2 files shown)


tests/e2e/helpers.go
```diff
@@ -1147,27 +1160,6 @@ func editFileOnNodes(f *framework.Framework, namespace string) error {
 	return nil
 }
 
-func addALottaFilesOnNodes(f *framework.Framework, namespace string) error {
```
rhmdnd, line 1150: IIUC - we were testing compression by adding a bunch of files to the host. Now we're testing compression another way?

mrogers950: Nope, it's the same method. I just converted what was a daemonset into a pod on a single node.

tests/e2e/helpers.go
```diff
@@ -467,13 +473,20 @@ func cleanAideAddedFilesDaemonset(namespace string) *appsv1.DaemonSet {
...
-// 20000 might not be enough
-// lowered to 10000
-func addLotsOfFilesDaemonset(namespace string) *appsv1.DaemonSet {
-	return privCommandDaemonset(namespace, "aide-add-lots-of-files",
-		"for i in `seq 1 10000`; do mktemp \"/hostroot/etc/addedbytest$i.XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX\"; done || true",
-	)
+// This pod is to stuff the aide log with added files, to force compression. The trick is in the long temp name that pads the aide log.
+// 10000 files.
+func addCompressionTestFilesOnNode(t *testing.T, f *framework.Framework, ctx *framework.Context, node, namespace string) error {
+	pod := privPodOnNode(namespace, "test-compression-pod", node, compressionFileCmd)
```
rhmdnd, line 479: Does the pod name need to be unique, or will it be run serially?

rhmdnd, line 479: Ok - the `compressionFileCmd` is what I was missing before. I kept looking for what created the files, but now I see that it's the command.

rhmdnd (commented):
Looks good - just one more question inline.

tests/e2e/helpers.go
mrogers950: nah, it's serial

mrogers950:
Madhu verified this fixed his issues with this test on Z

PR #258, opened 2022-07-07
Makefile: Add CATALOG_DEPLOY_NS variable for make catalog-deploy

mrogers950:
Setting CATALOG_DEPLOY_NS allows using catalog-deploy to a non-default
namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,9 @@ ifneq ($(origin CATALOG_BASE_IMG), undefined)
 FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
 endif
 
+# Set CATALOG_DEPLOY_NS= when running `make catalog-deploy` to override the default.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -141,6 +141,9 @@ ifneq ($(origin CATALOG_BASE_IMG), undefined)
 FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
 endif
 
+# Set CATALOG_DEPLOY_NS= when running `make catalog-deploy` to override the default.
+CATALOG_DEPLOY_NS ?= $(NAMESPACE)
```
rhmdnd, line 145: Do we need to document the fix for this in the CHANGELOG.md?

mrogers950: Yeah, that should be taken care of by the changelog script

PR #259, opened 2022-07-07
Bug 2104897: Use ClusterRole and local RoleBinding for daemon permissions

mrogers950:
Fixes permissions for an OLM install when installing into a namespace other
than openshift-file-integrity.

- Remove the daemon RoleBinding from kustomize under config/rbac. This avoids
  adding an openshift-file-integrity namespace to the subject SA during bundle
  creation.
- Change the daemon Role to a ClusterRole

ref: https://bugzilla.redhat.com/show_bug.cgi?id=2104897

### DIFF
--- a/config/rbac/daemon_rolebinding.yaml
+++ b/config/rbac/daemon_rolebinding.yaml
@@ -1,3 +1,6 @@
+# Daemon rolebinding is not applied to bundle/manifests by bundle generation on purpose, instead we should copy by hand
+#  to bundle/manifests. Reason being to avoid adding a namespace to subjects.
+---
 kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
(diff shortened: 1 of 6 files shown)


config/rbac/daemon_rolebinding.yaml
```diff
@@ -1,3 +1,5 @@
+# Daemon rolebinding is not applied to bundle/manifests by bundle generation on purpose, instead we should copy by hand
+#  to bundle/manifests. Reason being to avoid adding a namespace to subjects.
```
rhmdnd, line 2: Does bundling automatically add namespaces to these resources?

mrogers950: Yeah, bundling does a kustomize + bundle generation, and kustomize is set up to add the namespace (which unfortunately for our case, includes the SA subject for a role).

rhmdnd:
I checked this out locally and it worked for me. Waiting for a clean CI run to kick it through.

rhmdnd (approved):
/lgtm

Thanks, Matt!

PR #261, opened 2022-07-12
Makefile: Fix build dir recipe override

mrogers950:
Since the default build directory is "build", the $(BUILD_DIR) target
triggered an override warning.

  Makefile:310: warning: overriding recipe for target 'build'                                                                                                                                                                            
  Makefile:238: warning: ignoring old recipe for target 'build'

This removes the $(BUILD_DIR) target since ./build is part of the git
tree and should not need to be created prior to other targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,10 +233,6 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
-
-$(BUILD_DIR):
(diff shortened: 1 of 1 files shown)


rhmdnd (approved):
Checked this out locally and it cleans up the warnings.

Issue #260, opened 2022-07-08
Using *undeploy Makefile targets outputs Not Found errors


rhmdnd:
I was recently using the `make catalog-deploy` target to deploy changes into one of my clusters. When I needed to remove, or uninstall, the operator I used `make catalog-undeploy`, which outputs the following:

```
$ make catalog-undeploy
<snip>
namespace "openshift-file-integrity" deleted
customresourcedefinition.apiextensions.k8s.io "fileintegrities.fileintegrity.openshift.io" deleted
customresourcedefinition.apiextensions.k8s.io "fileintegritynodestatuses.fileintegrity.openshift.io" deleted
serviceaccount "file-integrity-daemon" deleted
serviceaccount "file-integrity-operator" deleted
role.rbac.authorization.k8s.io "leader-election-role" deleted
clusterrole.rbac.authorization.k8s.io "file-integrity-operator-metrics" deleted
clusterrole.rbac.authorization.k8s.io "fileintegrity-editor-role" deleted
clusterrole.rbac.authorization.k8s.io "fileintegrity-viewer-role" deleted
rolebinding.rbac.authorization.k8s.io "file-integrity-daemon" deleted
rolebinding.rbac.authorization.k8s.io "leader-election-rolebinding" deleted
rolebinding.rbac.authorization.k8s.io "prometheus-k8s" deleted
clusterrolebinding.rbac.authorization.k8s.io "file-integrity-operator-metrics" deleted
deployment.apps "file-integrity-operator" deleted
Error from server (NotFound): error when deleting "STDIN": roles.rbac.authorization.k8s.io "file-integrity-daemon" not found
Error from server (NotFound): error when deleting "STDIN": roles.rbac.authorization.k8s.io "file-integrity-operator" not found
Error from server (NotFound): error when deleting "STDIN": clusterroles.rbac.authorization.k8s.io "file-integrity-operator" not found
Error from server (NotFound): error when deleting "STDIN": rolebindings.rbac.authorization.k8s.io "file-integrity-operator" not found
Error from server (NotFound): error when deleting "STDIN": clusterrolebindings.rbac.authorization.k8s.io "file-integrity-operator" not found
make: *** [Makefile:357: undeploy] Error 1
```

It appears some of the resources were cleaned up properly, but others weren't. This may be confusing to contributors (do I need to go clean things up manually?)

PR #263, opened 2022-07-12
Makefile: Make uninstall/undeploy targets more robust

mrogers950:
- Update config/ kustomizations to distinct default and default-bundle usages in order to fix the daemon RoleBinding installation for `make deploy`, and to avoid substitution for the bundle generation.
- `make catalog-undeploy` works with CATALOG_DEPLOY_NS and removes the catalog-installed CSV object to take the operator down.
- Use --ignore-not-found=true for deletes.

Fixes https://github.com/openshift/file-integrity-operator/issues/260
@rhmdnd @xiaojiey This fixes a normal `make deploy` which broke after 10f546c1eed34848e6c64b995c924502808591ec


rhmdnd (approved):
/lgtm

I ran this against one of my clusters and it works as expected.

Thanks for cleaning this up, Matt!

PR #264, opened 2022-07-12
Build setup-envtest from vendor

mrogers950:
Fixes https://github.com/openshift/file-integrity-operator/issues/251


rhmdnd:
Should we still track tool dependencies in `tools.go`?

PR #271, opened 2022-07-20
Ensure we build latest images when we release

rhmdnd:
We have a couple of final steps in the `release-images` target that
tags the latest images with the most recent release. But, to do this, we
need to build the images and tag them locally before we push them to
Quay. Not doing this will result in failures to release the images
because podman can't find the tag locally.

This commit fixes the issue by adding steps to build the images locally
and then tag them with latest before pushing to Quay.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -470,7 +470,9 @@ push-release: package-version-to-tag ## Do an official release (Requires permiss
 .PHONY: release-images
 release-images: package-version-to-tag push catalog
 	# This will ensure that we also push to the latest tag
+	$(MAKE) images TAG=latest
 	$(MAKE) push TAG=latest
+	$(MAKE) catalog-image TAG=latest
 	$(MAKE) catalog-push TAG=latest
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -470,7 +470,9 @@ push-release: package-version-to-tag ## Do an official release (Requires permiss
 .PHONY: release-images
 release-images: package-version-to-tag push catalog
 	# This will ensure that we also push to the latest tag
+	$(MAKE) images TAG=latest
 	$(MAKE) push TAG=latest
+	$(MAKE) catalog-image TAG=latest
```
rhmdnd, line 475: This is going to result in a new container build with a new ID. It would be better to build once and then tag with the version and the `latest` tag.

rhmdnd:
Closing this in favor of an approach that's more consistent with what we do in the compliance-operator.

https://github.com/openshift/file-integrity-operator/pull/290

PR #272, opened 2022-07-21
Bug 2109153: Restore CSV ownership of ServiceAccounts

mrogers950:
- Use the --extra-service-accounts flag when generating the bundle in order to include the file-integrity-daemon and file-integrity-operator-metrics SAs (and their roles) in the CSV. This maintains the structure of the pre-0.1.25 CSV that led to ownership of the SAs.
- Changes to config/kustomize to allow --extra-service-accounts to pick up the SAs.
- Delete config/rbac/role.yaml which was a duplicate and unused.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2109153


rhmdnd:
This worked for me locally upgrading from 0.1.24 to 0.1.29.

Makefile
```diff
@@ -107,6 +107,9 @@ IMAGE_TAG_BASE=$(IMAGE_REPO)/$(APP_NAME)
 # You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:$(TAG)
 
+# Includes additional service accounts into the bundle CSV.
```
rhmdnd, line 110: Will this bite us if we leave it here? Should we link to the bug report that shows why we have to do this?

config/default-bundle/kustomization.yaml
```diff
@@ -1,4 +1,4 @@
-# Bundle variant - no rbac-daemon inclusion
```
rhmdnd, line 1: This bit is still true, right?

rhmdnd (approved):
Only minor comments/questions. Otherwise, this works for me.

rhmdnd:
/retest

I think the failure might be stale, or from before the most recent commit.

```
helpers.go:265: failed to initialize cluster resources: ClusterRoleBinding.rbac.authorization.k8s.io "file-integrity-operator-metrics" is invalid: subjects[1].namespace: Required value
```

Makefile
mrogers950: I don't think so, mainly cause it's preserving our bundle layout from pre-upgrade

config/default-bundle/kustomization.yaml
mrogers950: No, because including the daemon SA with the bundle flag handles it correctly when dumping to CSV. This was needed when the role was in the manifest files and subject to namespace adding.

mrogers950:
> /retest
> 
> I think the failure might be stale, or from before the most recent commit.
> 
> ```
> helpers.go:265: failed to initialize cluster resources: ClusterRoleBinding.rbac.authorization.k8s.io "file-integrity-operator-metrics" is invalid: subjects[1].namespace: Required value
> ```

The e2e deployment variant is a little busted now, trying to fix

mrogers950:
Updated again trying to sort out the metrics permissions. I've moved more things around. Noting these limitations I ran into while getting to this point:

Can't make a CRB without a namespace subject:
```
      message: 'error creating clusterrolebinding file-integrity-operator-metrics:                                                                                                                       
        ClusterRoleBinding.rbac.authorization.k8s.io "file-integrity-operator-metrics"                                                                                                                   
        is invalid: subjects[1].namespace: Required value'                                                                                                                                               
      reason: InstallComponentFailed                                                                                                                                                                     
      status: "False"                                                                                                                                                                                    
      type: Installed                                                                                                                                                                                    
    message: 'error creating clusterrolebinding file-integrity-operator-metrics: ClusterRoleBinding.rbac.authorization.k8s.io                                                                            
      "file-integrity-operator-metrics" is invalid: subjects[1].namespace: Required                                                                                                                      
      value'                                                                                                                                                                                             

```

Can't make a Role with non-resource URLS.
```
    catalogSources: []                                                                                                                                                                                   
    message: 'retrying execution due to error: error creating role file-integrity-operator-metrics:                                                                                                      
      Role.rbac.authorization.k8s.io "file-integrity-operator-metrics" is invalid:                                                                                                                       
      rules[2].nonResourceURLs: Invalid value: []string{"/metrics", "/metrics-fio"}:                                                                                                                     
      namespaced rules cannot apply to non-resource URLs'                                                                                                                                                
    phase: Installing                                                                                                                                                                                    

```

mrogers950:
It was a cluster setup failure.
/retest

rhmdnd:
/lgtm
/approve

I was able to run this locally and upgrade from 0.1.24 to 0.1.29 successfully. Thanks for unwinding all this, Matt.

PR #277, opened 2022-08-01
Bug 2112394: Update PrometheusRule on operator startup

mrogers950:
Updates the PrometheusRule if it already exists and differs from the
in-code default. This ensures we always apply the default to bring in
any changes during an operator upgrade.

Also adds `get` and `update` permissions for PrometheusRules to the
operator role.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=2112394


cmd/manager/operator.go
```diff
@@ -291,9 +292,26 @@ func createIntegrityFailureAlert(ctx context.Context, client *monclientv1.Monito
 				},
 			},
 		},
-	}, metav1.CreateOptions{})
+	}
+	_, createErr := client.PrometheusRules(namespace).Create(ctx, promRule, metav1.CreateOptions{})
 	if createErr != nil && !kerr.IsAlreadyExists(createErr) {
-		log.Info("could not create prometheus rule for alert", createErr)
```
rhmdnd, line 296: Do we still want this layer of logging? Or does the caller handle the different error cases accordingly and log?

cmd/manager/operator.go
```diff
@@ -291,9 +292,26 @@ func createIntegrityFailureAlert(ctx context.Context, client *monclientv1.Monito
...
+			metav1.GetOptions{})
+		if getErr != nil {
+			return getErr
+		}
+		if !reflect.DeepEqual(currentPromRule.Spec, promRule.Spec) {
+			promRuleCopy := currentPromRule.DeepCopy()
+			promRuleCopy.Spec = promRule.Spec
+			if _, updateErr := client.PrometheusRules(namespace).Update(ctx, promRuleCopy,
+				metav1.UpdateOptions{}); updateErr != nil {
+				return updateErr
```
rhmdnd, line 312: Would an upgrade test from 0.1.24 to 0.1.30 be the only way to test for this?

cmd/manager/operator.go
mrogers950: This was basically a warning that someone would have to grep out. I'd prefer we error and have the caller log now.

cmd/manager/operator.go
mrogers950: I think it would better fit as a unit test..

mrogers950:
Added a unit test (this brings in ginkgo/gomega)

cmd/manager/operator_test.go
```diff
@@ -0,0 +1,53 @@
...
+				err := createIntegrityFailureAlert(ctx, fakeClient.MonitoringV1(), ns)
+				Expect(err).To(BeNil())
+				createdRule, err := fakeClient.MonitoringV1().PrometheusRules(ns).Get(ctx, defaultPrometheusAlertName, v1.GetOptions{})
+				Expect(err).To(BeNil())
+				Expect(createdRule).To(Not(BeNil()))
+				defaultRule := defaultPrometheusRule(defaultPrometheusAlertName, ns)
+				Expect(defaultRule).To(Not(BeNil()))
+				Expect(createdRule.Spec).To(BeEquivalentTo(defaultRule.Spec))
+			})
+			It("Updates the default Rule if it exists (and differs)", func() {
```
rhmdnd, line 34: Is it safe to assume that this rule is used exclusively by FIO?

rhmdnd:
Works for me locally - just one last question on if changing the permissions would have any ramifications for things other than FIO.

cmd/manager/operator_test.go
mrogers950: That's right

mrogers950:
> Works for me locally - just one last question on if changing the permissions would have any ramifications for things other than FIO.

No, that's just the required perms for the update part of this function.

rhmdnd (approved):
/lgtm

PR #279, opened 2022-08-19
e2e: Add PSP labels to test namespace

mrogers950:
This should fix issues with the test namespace regarding PSP changes.

### DIFF
--- a/tests/framework/resource.go
+++ b/tests/framework/resource.go
@@ -107,7 +107,16 @@ func (ctx *Context) getNamespace(ns string) (string, error) {
 	}
 	// create namespace
 	ns = ctx.GetID()
-	namespaceObj := &core.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}}
+	namespaceObj := &core.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: ns,
+			Labels: map[string]string{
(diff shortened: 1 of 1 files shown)


mrogers950:
@rhmdnd @jhrozek PTAL
Note: The bundle jobs are broken, and merging this should help the pj-rehearse runs for the fix

tests/framework/resource.go
```diff
@@ -107,7 +107,16 @@ func (ctx *Context) getNamespace(ns string) (string, error) {
 	}
 	// create namespace
 	ns = ctx.GetID()
-	namespaceObj := &core.Namespace{ObjectMeta: metav1.ObjectMeta{Name: ns}}
+	namespaceObj := &core.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: ns,
+			Labels: map[string]string{
+				"pod-security.kubernetes.io/enforce":             "privileged",
```
jhrozek, line 114: These are fine and I'm going to ack the patch, but without also setting `pod-security.kubernetes.io/warn: privileged` and `pod-security.kubernetes.io/audit: privileged` you might be getting warnigns and audit lines when the privileged pods are spawned. I don't think it matters much for test pods, but in case e.g. QE starts complaining, we know what to fix.

jhrozek (approved):
/lgtm

PR #280, opened 2022-08-22
Generate bundle

mrogers950:
This is to include the permission updates from be830f7 into the CSV. Normally we do this during the release; doing it before that makes sure these permissions are included for the bundle-ci jobs. 

ref: https://github.com/openshift/release/pull/31464
@rhmdnd @jhrozek PTAL. We'll override the bundle jobs on this PR as well.


jhrozek (approved):
/lgtm

PR #281, opened 2022-09-14
Updates AIDE Daemonset default tolerations to account for ROSA infra nodes

tonytheleg:
This PR updates the default daemonset toleration to also include infra nodes.

With FedRAMP clusters using ROSA exclusively, we noticed that the AIDE daemonset pods were not spinning up on infra nodes. Looks like the default setting for tolerations is only configured for master nodes, but ROSA also has a similar taint configured on infra nodes as well. 

```shell
$ oc describe nodes -l node-role.kubernetes.io/infra | grep Taint
Taints:             node-role.kubernetes.io/infra:NoSchedule
Taints:             node-role.kubernetes.io/infra:NoSchedule
```
We've fixed our issue by adding it to the `FileIntegrity` CR we deploy via OLM but I thought it might be good to update the default for other customers who may use FIO outside of FedRAMP. Chances are if a customer is leveraging FIO for their cluster, they are trying to meet some form of compliance requirement and will need all nodes scanned. The default should account for that in case they do not catch the config options available in the `FileIntegrity` CR

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ status:
 ```
 In the `spec`:
 * **nodeSelector**: Selector for nodes to schedule the scan instances on.
-* **tolerations**: Specify tolerations to schedule on nodes with custom taints. When not specified, a default toleration allowing running on master nodes is applied.
+* **tolerations**: Specify tolerations to schedule on nodes with custom taints. When not specified, a default toleration allowing running on master and infra nodes is applied.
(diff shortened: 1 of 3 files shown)


mrogers950:
@tonytheleg Thanks!
/approved
/lgtm

rhmdnd (approved):
Looks good. We'll add a release note in a subsequent patch.

PR #285, opened 2022-10-04
make: Workaround to make the upstram catalog usable on 4.12+

jhrozek:
NOTE: I'm not sure we should merge this PR. I'm mostly posting it up for
discussion. We could as well convert to a file based catalog:
```
opm migrate <registry_image> <fbc_directory>
opm generate dockerfile <fbc_directory> \
  --binary-image \
    registry.redhat.io/openshift4/ose-operator-registry:v4.11
```
or just create the file-based catalog from scratch. This PR might be useful
as a workaround though.

Implements the workaround found in:
    https://access.redhat.com/articles/6977554

This makes the FIO catalog usable on 4.12+. Note that fetching the
`registry.redhat.io/openshift4/ose-operator-registry:v4.11` image
requires authentication.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,7 +317,7 @@ bundle-image: bundle ## Build the bundle image.
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-image
 catalog-image: opm ## Build a catalog image.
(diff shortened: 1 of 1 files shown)


mrogers950:
@jhrozek thanks! It seems like adding the migrate and dockerfile generate to the catalog target would be the way to go at first.

jhrozek:
yeah, I'll open a new PR, thanks for the second opinion

PR #286, opened 2022-10-10
make: Build a file-based catalog

jhrozek:
Since OLM is deprecating the SQLite based catalog we are currently
using, while at the same time, the SQLite based catalogs have issues
interacting with OpenShift clusters that implement the recent PSA
changes, let's switch to file-based catalog:
    https://olm.operatorframework.io/docs/reference/file-based-catalogs/

Most of the work is done by the "opm render" command except for some
metadata that is stored at catalog/preamble.json, for example the icon.

I'll point out some questions inline.


jhrozek:
/hold
I'd like a thorough review, so holding the PR explicitly to avoid merging when someone clicks "Approve" in the UI.

Makefile
```diff
@@ -317,7 +319,13 @@ bundle-image: bundle ## Build the bundle image.
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-image
 catalog-image: opm ## Build a catalog image.
-	$(OPM) index add --container-tool $(RUNTIME) --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
```
jhrozek, line 320: So one thing I don't know how to test explicitly is the semver mode. Should I build a catalog, then push an updated version and see if OLM updates? Do we add just one bundle to the catalog?
I'm not sure what's the equivalent of the semver mode for the file based catalog, the closest I was pointed to on Slack are [veneers](https://olm.operatorframework.io/docs/reference/veneers/)

catalog/preamble.json
```diff
@@ -0,0 +1,20 @@
+{
```
jhrozek, line 1: Is `[catalog/preamble.json](https://github.com/openshift/file-integrity-operator/pull/286/commits/7f4f0f728ff81b7209be1ca10d69a20a734334f7#diff-3dd0e143b145e1a7ddf92660165b93f2673f13cdecf0d379d081e383e4de8cd6)` a reasonable filename and location?

Makefile
jhrozek: btw what I tested was:
 - make a new bundle with a made-up version (I used 1.2.3)
 - push the bundle
 - make the catalog with the same version
 - push the catalog
(NOTE: OPM doesn't work with local container caches. You really must push the images to a registry)
 - install the catalog on an OCP 4.12 cluster
 - verify the versions all say 1.2.3:
```
oc get csv,ip,og,sub -nopenshift-file-integrity
NAME                                                                        DISPLAY                   VERSION   REPLACES   PHASE
clusterserviceversion.operators.coreos.com/file-integrity-operator.v1.2.3   File Integrity Operator   1.2.3                Succeeded

NAME                                             CSV                              APPROVAL    APPROVED
installplan.operators.coreos.com/install-vzv5g   file-integrity-operator.v1.2.3   Automatic   true

NAME                                                         AGE
operatorgroup.operators.coreos.com/file-integrity-operator   26m

NAME                                                            PACKAGE                   SOURCE                    CHANNEL
subscription.operators.coreos.com/file-integrity-operator-sub   file-integrity-operator   file-integrity-operator   alpha

```

mrogers950: @jhrozek we always just add one bundle per catalog. So, I think that makes the mode irrelevant. 
The install test sounds good. Trying an upgrade should be just a matter of making the next fake version and running the make targets over.

PR #290, opened 2022-10-17
Fix make release-images to properly tag `latest`

rhmdnd:
Previously, we'd only tag the latest version of the operator we're
releasing in the upstream images available through
quay.io/file-integrity-operator.

In addition to the version tag (e.g., 0.1.31), we can also tag the
latest build with the `latest` tag, making it easier for users to deploy
the latest version of the upstream operator.


mrogers950:
Thanks!
/lgtm

PR #291, opened 2022-10-21
Fix deployment on OCP 4.6

mrogers950:
Check against changing the Service's clusterIP, and retain the existing value as OCP 4.6 does not tolerate that. Later versions just ignore the field.

(same fix as https://github.com/ComplianceAsCode/compliance-operator/commit/404483b5b1e0c5519da5ea89c6d68baecd2b1140)


jhrozek (approved):
/lgtm
I haven't actually tested this, but verified that this is exactly the same patch as for CO.

mrogers950:
> /lgtm I haven't actually tested this, but verified that this is exactly the same patch as for CO.

Thanks! I tested this on a 4.6 cluster, and it's all good.

mrogers950:
(Noticed the issue for https://github.com/openshift/file-integrity-operator/pull/292 while testing 4.6.)

PR #296, opened 2022-11-01
Ensure we update the ocp-0.1 branch when releasing

rhmdnd:
Some downstream machinery depends on the ocp-0.1 branch. Let's make sure
we sync the branch with the latest release during the `push-release`
process.

This is consistent with how we maintain the `ocp-0.1` branch in the
Compliance Operator.


jhrozek:
```
msg=Error: reading EC2 EIP (eipalloc-00ff00581a3391e86): couldn't find resource
```

PR #300, opened 2022-11-22
Release v1.0.0

rhmdnd:


bundle/manifests/file-integrity-operator.clusterserviceversion.yaml
```diff
@@ -19,10 +19,12 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: Monitoring,Security
-    olm.skipRange: '>=0.1.6 <0.1.32'
+    olm.skipRange: '>=0.1.6 <1.0.0'
```
rhmdnd, line 22: Note for other reviewers.

This allows someone to upgrade directly to 1.0.0, which is fine since 1.0.0 is technically backwards compatible with 0.1.32. Even though 1.0.0 is technically a major version bump, it doesn't change how it's used or deployed so a direct upgrade should be safe.

If or when we release FIO 2.0.0, we'll need to update the skipRange so that 1.0.0 is the lower bound, but that will be a ways out (we don't actually now when that would happen). We may not be supporting any 0.1.* releases at that point.

mrogers950:
I'm expecting `e2e-bundle-aws-upgrade` to fail.

rhmdnd:
Manually updated the upgrade versions here https://github.com/openshift/release/pull/34279 but there are still some issues with the upgrade testing scenario.

We'll have to merge this through like we did with 0.1.32.

PR #307, opened 2023-03-02
Bump golang dependency to 1.19

rhmdnd:
We were using 1.16, which ended support 11 months ago. Let's update the
version of go we use and also update the dependencies we use (fixing
vulnerabilities caught by dependabot).

You can verify the contents of this commit using:

  $ make clean // bump go version in go.mod
  $ go get -u all  // fix resolution issues manually
  $ go mod go_mod
  $ go mod vendor


rhmdnd:
Bumping golang version for CI here - https://github.com/openshift/release/pull/36861

rhmdnd:
> so it looks like we also need to update `Dockerfile.ci` and tell it to use `rhel-8-release-golang-1.19-openshift-4.12` instead?

Aha - good call. I forgot about that hook.

Thanks!

jhrozek (approved):
/lgtm

PR #308, opened 2023-03-02
Configure dependabot

jhrozek:
We configure dependabot for FIO because FIO is hosted under the
openshift org and renovate does seem to require org admin access to be
set up and in general I'm uneasy about configuring an app in the
openshift org.


jhrozek:
ha, `500 Internal Server Error` in CI

rhmdnd (approved):

jhrozek:
looks like a test timeout

PR #312, opened 2023-03-06
Release v1.2.0

rhmdnd:


bundle/manifests/file-integrity-operator.clusterserviceversion.yaml
```diff
@@ -320,4 +321,7 @@ spec:
   provider:
     name: Red Hat
     url: https://github.com/openshift/file-integrity-operator
-  version: 1.0.0
+  relatedImages:
+  - image: quay.io/file-integrity-operator/file-integrity-operator:latest
```
rhmdnd, line 325: I don't think this should be in here? It could be that this is generated from newer versions of operator-sdk, but I don't think it should be pointing to latest.

jhrozek (approved):
/lgtm

PR #314, opened 2023-03-06
Use remote detection in release process

rhmdnd:
Some contributors use different naming conventions and protocols for
their upstream and downstream branchs. Instead of assuming the same
setup for everyone, let's use a script to find which remote name to use
based on the organization and repository, and then call that script from
the release targets.

This makes the release process a little easier for contributors,
regardless of how they have their remotes setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,13 @@ endif
 
 # Git options.
 GIT_OPTS?=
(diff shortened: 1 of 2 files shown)


Vincent056 (approved):
/lgtm

rhmdnd:
Adding necessary labels since this is specific to the upstream release process.

PR #321, opened 2023-03-14
Update CI to use RHEL 9 and OpenShift 4.13

rhmdnd:
This changes accompanies CI changes to start using OpenShift 4.13 with
RHEL 9 base images.

  https://github.com/openshift/release/pull/37184

We want to start testing FIO using RHEL 9 because it relies on AIDE to
detect file system changes. We need to make sure we don't have any
regressions using RHEL 9.


rhmdnd:
/retest

Looks like a potential issue. Retesting to see if we get the same problem.

JAORMX (approved):

PR #368, opened 2023-05-30
OCPBUGS-14348 Change the default config file to exclude kubelet CA

Vincent056:
We want to exclude the Kubelet CA file from the default config, so the user does not have to reinit the aide database when the cert gets renewed.

Link to the Jira Issue: [OCPBUGS-14348](https://issues.redhat.com/browse/OCPBUGS-14348)


jhrozek (approved):
Since this is causing issues, I don't see any other way
/lgtm

PR #389, opened 2023-06-28
tests: Set logger to an empty sink in tests

jhrozek:
This is pretty much
https://github.com/fluxcd/flux2/pull/3932/commits/aa65589391db2433691d8cebde5c77c090cae3a9
reused.

The new controller-runtime version requires that the logger is set or
else there's a stack trace printed. Since even the original code doesn't
use any logger for tests, let's just use null logger with the new
version as well.

### DIFF
--- a/tests/framework/framework.go
+++ b/tests/framework/framework.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


rhmdnd:
/retest

The metrics must have been an issue with something in the underlying CI. I was unable to reproduce it locally and the result affected more that just this patch.

rhmdnd:
Rebased this on master to get a clean test run in https://github.com/openshift/file-integrity-operator/pull/421

PR #399, opened 2023-07-11
Update aide database configuration option

rhmdnd:
Newer versions of aide deprecated the `database` configuration option in
favor of `database_in`.

This commit updates the aide configuration we use so we don't emit a
deprecation warning in the aide logs.

### DIFF
--- a/aide.conf.rhel8
+++ b/aide.conf.rhel8
@@ -4,7 +4,7 @@
 @@define LOGDIR /var/log/aide
 
 # The location of the database to be read.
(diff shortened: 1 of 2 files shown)


rhmdnd:
We will need to test the upgrade path here to make sure the existing database works after FIO is upgraded.

rhmdnd:
Closing this for now since we pinned the version of AIDE to a version that works with FIO. We can reopen this when we do a more comprehensive update to support newer versions of AIDE.

PR #403, opened 2023-07-20
Add clean-images command to purge build env

yuumasato:
Add Make command to remove all unused containers, images, pods and networks.
Sometimes the build may end up using a cached stale image for no apparent reason.
This command facilitates purging images used to build the operator and pulling fresh images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,12 @@ else
 RUNTIME_BUILD_CMD=build
 endif
 
+ifeq ($(RUNTIME),buildah)
+RUNTIME_PRUNE_CMD=prune --all --force
+else
+RUNTIME_PRUNE_CMD=system prune --all --force
+endif
+
 ifeq ($(RUNTIME), podman)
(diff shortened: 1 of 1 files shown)


rhmdnd (approved):
/lgtm

Makefile
```diff
@@ -189,6 +195,10 @@ clean-kustomize: ## Reset kustomize changes in the repo.
 clean-controller-gen: ## Remove the controller-gen build
 	rm -f $(CONTROLLER_GEN)
 
+.PHONY: clean-images
```
rhmdnd, line 198: nit: we could use a name that more inclusive of image since we're cleaning up quite a bit. But we can change that in another patch, too.

yuumasato: @rhmdnd I agree, but I couldn't come up with a good name. Do you have any suggestion?

How about `clean-build-assets` or `clean-runtime-data`?

PR #434, opened 2023-09-12
OCPBUGS-18532: exclude file for ServiceMesh Operator

Vincent056:
we are excluding `/hostroot/etc/cni/multus/net.d` from the default aide.conf file so that ServiceMesh Operator can run without triggering FIO failures.

[OCPBUGS-18532](https://issues.redhat.com/browse/OCPBUGS-18532)

### DIFF
--- a/pkg/controller/fileintegrity/config_defaults.go
+++ b/pkg/controller/fileintegrity/config_defaults.go
@@ -52,6 +52,6 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/machine-config-daemon/node-annotation.json*
 !/hostroot/etc/pki/ca-trust/extracted/java/cacerts$
 !/hostroot/etc/cvo/updatepayloads
-
+!/hostroot/etc/cni/multus/net.d/*
 # Catch everything else in /etc
 /hostroot/etc/    CONTENT_EX`
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -165,6 +165,31 @@ func TestMetricsHTTPVersion(t *testing.T) {
 	t.Log("Metrics endpoints use HTTP/1.1, as expected")
 }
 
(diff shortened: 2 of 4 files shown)


rhmdnd:
We should be good to rebase this one now that we landed the CI fix in https://github.com/openshift/file-integrity-operator/pull/440

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -52,6 +52,6 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/machine-config-daemon/node-annotation.json*
 !/hostroot/etc/pki/ca-trust/extracted/java/cacerts$
 !/hostroot/etc/cvo/updatepayloads
-
+!/hostroot/etc/cni/multus/net.d/*
```
rhmdnd, line 55: To test this in CI, we'd need to setup a cluster with a CNI plugin to make sure this is excluded, right?

rhmdnd (commented):
Change looks fine, just wondering how we would test this against regressions.

pkg/controller/fileintegrity/config_defaults.go
Vincent056: that's correct

Vincent056:
> Change looks fine, just wondering how we would test this against regressions.

I think the config will be tested for all regular FIO operation in the CI, however, we need to manually test if this issue has been fixed the ServiceMesh operator issue, let me check if there a automated way doing this

rhmdnd:
Looks like the e2e tests failed because of a pre-existing service account:

```
     wait_util.go:59: Deployment available (1/1)
    helpers.go:591: Failed to initialize cluster resources for ServiceMeshOperator: serviceaccounts "file-integrity-daemon" already exists 
```

xiaojiey:
Details seen from https://issues.redhat.com/browse/OCPBUGS-18532

Vincent056:
Rebase to resolve conflict

Issue #439, opened 2023-09-16
E2E testing fails due to failed node taints


rhmdnd:
We're experiencing an issue in CI where the E2E testing fails because a node can't be tainted, which manifests in the follow error:

```
 === RUN   TestFileIntegrityTolerations
&Namespace{ObjectMeta:{osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[pod-security.kubernetes.io/enforce:privileged security.openshift.io/scc.podSecurityLabelSync:false] map[] [] [] []},Spec:NamespaceSpec{Finalizers:[],},Status:NamespaceStatus{Phase:,Conditions:[]NamespaceCondition{},},}
    client.go:47: resource type ServiceAccount with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-daemon) created
    client.go:47: resource type ServiceAccount with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-operator) created
    client.go:47: resource type Role with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-daemon) created
    client.go:47: resource type Role with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-operator) created
    client.go:47: resource type Role with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/leader-election-role) created
    client.go:47: resource type ClusterRole with namespace/name (/file-integrity-operator) created
    client.go:47: resource type ClusterRole with namespace/name (/file-integrity-operator-metrics) created
    client.go:47: resource type ClusterRole with namespace/name (/fileintegrity-editor-role) created
    client.go:47: resource type ClusterRole with namespace/name (/fileintegrity-viewer-role) created
    client.go:47: resource type RoleBinding with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-daemon) created
    client.go:47: resource type RoleBinding with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-operator) created
    client.go:47: resource type RoleBinding with namespace/name (osdk-e2e-6c36263d-255d-4a22-9659-ba12e7965054/file-integrity-operator-metrics) created
[...]

Issue #438, opened 2023-09-16
E2E testing is broken waiting for nodes to start


rhmdnd:
Seeing this consistently across various patches (e.g., dependency bumps).

https://github.com/openshift/file-integrity-operator/pull/432

```
     wait_util.go:59: Deployment available (1/1)
    client.go:47: resource type  with namespace/name (osdk-e2e-0c35dce2-3be6-43fd-9624-566a1176cb0f/e2e-test-certrotation) created
    helpers.go:393: Created FileIntegrity: &{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:e2e-test-certrotation GenerateName: Namespace:osdk-e2e-0c35dce2-3be6-43fd-9624-566a1176cb0f SelfLink: UID:576fd4be-282f-4198-9c13-135c08c2818f ResourceVersion:62749 Generation:1 CreationTimestamp:2023-09-06 21:45:15 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[] Annotations:map[] OwnerReferences:[] Finalizers:[] ManagedFields:[{Manager:Go-http-client Operation:Update APIVersion:fileintegrity.openshift.io/v1alpha1 Time:2023-09-06 21:45:15 +0000 UTC FieldsType:FieldsV1 FieldsV1:{"f:spec":{".":{},"f:config":{".":{},"f:gracePeriod":{},"f:maxBackups":{}},"f:debug":{},"f:nodeSelector":{".":{},"f:node-role.kubernetes.io/worker":{}},"f:tolerations":{}}} Subresource:}]} Spec:{NodeSelector:map[node-role.kubernetes.io/worker:] Config:{Name: Namespace: Key: GracePeriod:20 MaxBackups:5 InitialDelay:0} Debug:true Tolerations:[{Key:node-role.kubernetes.io/master Operator:Exists Value: Effect:NoSchedule TolerationSeconds:<nil>} {Key:node-role.kubernetes.io/infra Operator:Exists Value: Effect:NoSchedule TolerationSeconds:<nil>}]} Status:{Phase:}}
    helpers.go:902: Got (Active) result #1 out of 0 needed.
    helpers.go:913: FileIntegrity ready (Active)
    helpers.go:430: FileIntegrity deployed successfully
    helpers.go:902: Got (Active) result #1 out of 0 needed.
    helpers.go:913: FileIntegrity ready (Active)
[...]

PR #440, opened 2023-09-19
Fix TestFileIntegrityCertRotation e2e test

Vincent056:
Change the how we check if a kubelet cert rotation is being completed, and make sure FIO object pass after the rotation

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -22,8 +22,8 @@ import (
 	"github.com/openshift/file-integrity-operator/pkg/controller/metrics"
 	"github.com/pborman/uuid"
 
+	configv1 "github.com/openshift/api/config/v1"
 	v1 "k8s.io/api/rbac/v1"
-
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 
@@ -40,10 +40,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kerr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/file-integrity-operator/pkg/common"
 )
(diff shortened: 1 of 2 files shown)


rhmdnd:
Is this related to either of the open issues we have for broken e2e tests?

https://github.com/openshift/file-integrity-operator/issues/438
https://github.com/openshift/file-integrity-operator/issues/439

rhmdnd:
/retest

Looks like the e2e setup failed due to `ImagePullBackoff` (seemed unrelated).

Vincent056:
failed due to ImagePullBackoff

Vincent056:
TestFileIntegrityCertRotation is fixed, however there are issue with node taint on bundle e2e test

tests/e2e/helpers.go
```diff
@@ -1024,27 +1225,101 @@ func assertNodeOKStatusEvents(t *testing.T, f *framework.Framework, namespace st
...
+	err := createFromYAML(t, f, []byte(certRotationYaml), nil)
 	if err != nil {
-		t.Error("Failed to get secret for kube-apiserver-to-kubelet-signer, err: ", err)
+		t.Error(err)
 		return
 	}
-	if secret.Annotations == nil {
-		secret.Annotations = map[string]string{}
+	// wait for cert rotation daemonset to be ready
+	err = waitForDaemonSet(daemonSetIsReady(f.KubeClient, "kubelet-bootstrap-cred-manager", "openshift-machine-config-operator"))
```
rhmdnd, line 1267: nit: Potential area for improvement (although I don't expect us to do it in this patch), would be to consider a simplified polling method here, instead of a nest one:

```go
  name := "kubelet-bootstrap-cred-manager"
  namespace := "openshift-machine-config-operator"
  err = assertDaemonSetIsActive(name, namespace)
```

tests/e2e/helpers.go
```diff
@@ -1024,27 +1225,101 @@ func assertNodeOKStatusEvents(t *testing.T, f *framework.Framework, namespace st
...
+	if err != nil {
+		t.Errorf("Timed out waiting for DaemonSet kubelet-bootstrap-cred-manager")
+	}
+
+	// Delete the secrets csr-signer-signer and csr-signer from the openshift-kube-controller-manager-operator namespace
```
rhmdnd, line 1272: We do this because it's what forces the rotation, right?

tests/e2e/helpers.go
```diff
@@ -1755,38 +2030,45 @@ func waitUntilPodsAreGone(t *testing.T, c client.Client, pods *corev1.PodList, i
 }
 
 func taintNode(t *testing.T, f *framework.Framework, node *corev1.Node, taint corev1.Taint) error {
-	taintedNode := node.DeepCopy()
-	if taintedNode.Spec.Taints == nil {
-		taintedNode.Spec.Taints = []corev1.Taint{}
-	}
-	taintedNode.Spec.Taints = append(taintedNode.Spec.Taints, taint)
-	t.Logf("Tainting node: %s", taintedNode.Name)
-
```
rhmdnd, line 1764: Interesting - so what this causing issues in previous CI runs because it wasn't being retried? Or we're working around the CI issue because we're retrying?

tests/e2e/helpers.go
```diff
@@ -1755,38 +2030,45 @@ func waitUntilPodsAreGone(t *testing.T, c client.Client, pods *corev1.PodList, i
 }
 
 func taintNode(t *testing.T, f *framework.Framework, node *corev1.Node, taint corev1.Taint) error {
```
rhmdnd, line 2032: nit: Taints are specific to nodes, we could simplify the name to `addTaint`.

tests/e2e/helpers.go
```diff
@@ -1755,38 +2030,45 @@ func waitUntilPodsAreGone(t *testing.T, c client.Client, pods *corev1.PodList, i
...
+				taintedNode.Spec.Taints = []corev1.Taint{}
+			}
+			taintedNode.Spec.Taints = append(taintedNode.Spec.Taints, taint)
+			t.Logf("Tainting node: %s", taintedNode.Name)
 			return f.Client.Update(goctx.TODO(), taintedNode)
 		},
 	)
 }
 
 func removeNodeTaint(t *testing.T, f *framework.Framework, nodeName, taintKey string) error {
```
rhmdnd, line 2054: nit: Similar comment as above, we could simplify to `removeTaint`

tests/e2e/helpers.go
```diff
@@ -1755,38 +2030,45 @@ func waitUntilPodsAreGone(t *testing.T, c client.Client, pods *corev1.PodList, i
...
-		taintedNode.Spec.Taints = []corev1.Taint{}
-	}
-	taintedNode.Spec.Taints = append(taintedNode.Spec.Taints, taint)
-	t.Logf("Tainting node: %s", taintedNode.Name)
-
 	return retryDefault(
 		func() error {
+			// taint with retry
+			// let's fetch the latest node object first
+			fetchedNode := &corev1.Node{}
```
rhmdnd, line 2037: Do we need to fetch the node again, even though it's passed in, because we're not sure if it has been updated previously and cause our update to fail because we're attempting to update an older version?

rhmdnd (approved):
I only have minor suggestions inline, most of which we can address in subsequent patches if we decide to do so.

Thanks for fixing up the CI, @Vincent056

PR #454, opened 2023-10-17
Use golang 1.20

rhmdnd:
Let's us a version of golang that contains a patched version of
net/http.

https://github.com/golang/go/issues/63426


yuumasato (approved):
/lgtm

PR #455, opened 2023-10-18
Disable http2 as default for webhook

Vincent056:
In order to address CVE-2023-44487, we need to disable the http2 by default in webhook.


rhmdnd (approved):
/lgtm

PR #469, opened 2023-10-27
Implement build flags

rhmdnd:
Implement an environment variable that we can use to pass additional
arguments into the build process. This is useful for folks that want to
build the operator in specific ways (e.g., FIPS enabled).


Vincent056 (approved):
/lgtm

PR #524, opened 2024-05-10
Use ubi images instead of fedora minimal

rhmdnd:
Fedora repositories are using newer versions of AIDE that we don't
support with FIO, which will cause issues running the operator because
it doesn't know how to configure AIDE 0.18.

This commit updates the dockerfiles to use UBI images which provide AIDE
0.16, which is compatible with FIO and gives us some time to implement
support for 0.18.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -11,8 +11,7 @@ COPY . .
 RUN make build
 
 # Step two: containerize file-integrity-operator and AIDE together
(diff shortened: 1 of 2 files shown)


rhmdnd:
Trying an alternative here - https://github.com/openshift/file-integrity-operator/pull/526

I don't think the UBI images have aide in any of there repositories, which is probably why we were using Fedora in the first place.

PR #525, opened 2024-05-10
CMP-2175: Implement support for replaces attribute in CSV

rhmdnd:
Doing this makes it so that OLM can build dependency graphs between
operator versions, which make it easier for users perform installs on
disconnected environments and upgrades.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ GIT_OPTS?=
 # We rely on a bash script for this since it's simplier than interating over a
 # list with conditionals in GNU make.
 GIT_REMOTE?=$(shell ./utils/git-remote.sh)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -505,3 +508,7 @@ release-images: package-version-to-tag push catalog
 .PHONY: changelog
 changelog:
 	@utils/update_changelog.sh "$(TAG)"
+
+.PHONY: set-current-version
+set-current-version:
+	@echo "Using $(PREVIOUS_VERSION) as previous version"
```
rhmdnd, line 514: This outputs the old version, but when the actual substitution takes place it puts the new version in. Need to investigate that.

Vincent056:
@rhmdnd could you rebase this PR, thanks!

PR #526, opened 2024-05-10
Use Fedora 37 for AIDE 0.16

rhmdnd:
Newer versions of Fedora have AIDE 0.18, which breaks with FIO due to
backwards incompatible configuration changes with AIDE 0.18.

This commit uses an older version of fedora so we can still build FIO
until integrate a newer version of AIDE.

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -11,7 +11,7 @@ COPY . .
 RUN make build
 
 # Step two: containerize file-integrity-operator and AIDE together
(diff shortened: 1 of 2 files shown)


Vincent056 (approved):
/lgtm

rhmdnd:
Looks like the upgrade testing is failing on metrics assertion issues.

yuumasato:
We can stay for a while on F37, but it is EOL since Dec 2023.
https://docs.fedoraproject.org/en-US/releases/eol/

PR #528, opened 2024-05-13
Update manifests for kubebuilder version

rhmdnd:
Makefile tooling is generating manifests with older kubebuilder
versions. Let's just update these so they're up-to-date with the version
we're using.


Vincent056 (approved):
/lgtm

PR #534, opened 2024-05-29
OCPBUGS-31257: Exclude additional files in default aide conf

Vincent056:
We are going to exlucde following in the default aide conf: 
`!/hostroot/etc/cni/multus/certs` for the OVN-Kubernetes CNI
`!/hostroot/etc/kubernetes/compliance-operator` for the Compliance Operator check runtime kubeletconfig `!/hostroot/etc/kubernetes/node-feature-discovery` for the Node Feature Discovery Operator

https://issues.redhat.com/browse/OCPBUGS-31257


rhmdnd (approved):
/lgtm

Seems reasonable and existing e2e tests will verify the config works with AIDE.

It would be nice to have a separate set of situational tests that exercise the excluded config with the necessary operator, but that's a larger lift than what I think these e2e tests were intended for.

PR #537, opened 2024-07-02
AIDE migration

Vincent056:
This PR adds the ability to run aide 0.18 config checks and migration.

The aide0.18 will be compiled as an additional binary and used by FIO to check our migrated config. This new binary will reside in the container image, and we will launch a new pod to check the aide config migration every time we need to re-init. 

Add config check to existing, upgrade case, check if migration checks needed annotation, if so we launch the pod, pod checks for aide migration and issue warning return result as a config map, the config map controller pick up the change, remove migration checks needed annotation.

We will try to perform the migration if a user-defined config is detected. We will issue warning messages in the log, and have a failed annotation key in the FIO instance if we cannot pre-migrate the config.

 Annotation keys added:

```
	// AideConfigMigrationIgnoreAnnotationKey tells us to ignore the deprecated config options
	AideConfigMigrationIgnoreAnnotationKey = "file-integrity.openshift.io/migration-ignore-deprecation"
	// AideConfigMigrationFailedAnnotationKey tells us that the migration failed
	AideConfigAutoMigrationFailedAnnotationKey = "file-integrity.openshift.io/migration-failed"
	// AideConfigMigrationCheckDisabledAnnotationKey tells us that the migration check is disabled
	AideConfigMigrationCheckDisabledAnnotationKey = "file-integrity.openshift.io/migration-check-disabled"
```

example of event message:

```
7m17s       Warning   FileIntegrityAIDEConfigMigration   fileintegrity/example-fileintegrity            Migration check failed: Detected error config during the migration check: Invalid configureline error
```

## AIDE Configuration Changes
 - Only contains things that are being deprecated or removed along with their potential replacement

### Removed Features in AIDE v0.17

- **ignore_list**
  - Attributes whose changes are ignored in the report.

- **report_attributes**
  - Attributes always printed in the report for changed files. If an attribute is both ignored and forced, it is not considered for file change but will be printed in the final report if the file has otherwise changed.

- **verbose** (type: number, range: 0 - 255, default: 5)
  - Removed, use `log_level` and `report_level` options instead.

### New Features in AIDE v0.17

- **database_in** (type: URL, default: see \fB--version\fP output, added in AIDE v0.17)
- **log_level** (type: log level, default: \fBwarning\fR)
- **report_level** (type: report_level, default: \fBchanged_attributes\fR)
  - Specifies the log level. Log messages are written to \fIstderr\fR. If there are multiple \fIlog_level\fR lines, the first one is used. The `--log-level` or `-L` command line option overwrites this option.

### New Features in AIDE v0.16

- **report_ignore_added_attrs** (type: attribute expression, default: empty)
  - Attributes whose addition is ignored in the report.

- **report_ignore_removed_attrs** (type: attribute expression, default: empty)
  - Attributes whose removal is ignored in the report.

- **report_ignore_e2fsattrs** (type: string, default: 0)
  - List (no delimiter) of ext2 file attributes to be ignored in the report.

- **report_force_attrs** (type: attribute expression, default: empty)
  - Attributes always printed in the report for changed files.

### Deprecated Features in AIDE v0.18 (to be removed in AIDE v0.20)

- **@@ifdef VARIABLE**
  - Same as `@@if defined VARIABLE`.

- **@@ifndef VARIABLE**
  - Same as `@@if not defined VARIABLE`.

- **@@ifhost HOSTNAME**
  - Same as `@@if hostname HOSTNAME`.

- **@@ifnhost HOSTNAME**
  - Same as `@@if not hostname HOSTNAME`.

- **Special attributes**
  - **S**
    - Check for growing size. (DEPRECATED since AIDE v0.18, will be removed in AIDE v0.20)
    - Use `growing+s` attributes instead.

### Removed Features in AIDE v0.19

- **database**
  - The URL from which the database is read. Only one of these lines is allowed. If there are multiple database lines, the first is used.


pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -55,6 +55,7 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/cni/multus/certs
 !/hostroot/etc/kubernetes/compliance-operator
 !/hostroot/etc/kubernetes/node-feature-discovery
+!/hostroot/etc/mco/internal-registry-pull-secret.json$
```
rhmdnd, line 58: Interesting.. I wonder if I'm seeing e2e error because of this, too.

rhmdnd: Pulled this out into another change to see if we can get a clean run in isolation.

https://github.com/openshift/file-integrity-operator/pull/541

Vincent056:
we will save migrate config in:

```
kind: ConfigMap
apiVersion: v1
metadata:
  name: example-fileintegrity
  namespace: openshift-file-integrity
  uid: 3e68694f-fed1-403f-805f-813eec85ff38
  resourceVersion: '68362036'
  creationTimestamp: '2024-06-28T06:22:38Z'
  labels:
    file-integrity.openshift.io/aide-conf: ''
    file-integrity.openshift.io/owner: example-fileintegrity
  annotations:
    kubernetes.io/description: hi
  managedFields:
    - manager: Mozilla
      operation: Update
      apiVersion: v1
      time: '2024-07-12T06:37:37Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:metadata':
          'f:annotations':
            .: {}
            'f:kubernetes.io/description': {}
    - manager: file-integrity-operator
      operation: Update
      apiVersion: v1
      time: '2024-07-12T18:35:21Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:data':
          .: {}
          'f:aide-0.18.conf': {}
          'f:aide.conf': {}
        'f:metadata':
          'f:labels':
            .: {}
            'f:file-integrity.openshift.io/aide-conf': {}
            'f:file-integrity.openshift.io/owner': {}
data:
  aide-0.18.conf: |-
    @@define DBDIR /hostroot/etc/kubernetes
    @@define LOGDIR /hostroot/etc/kubernetes
    database_in=file:@@{DBDIR}/aide.db.gz
    database_out=file:@@{DBDIR}/aide.db.gz.new
    gzip_dbout=yes
    log_level=notice
    report_level=added_removed_attributes
    report_url=file:@@{LOGDIR}/aide.log.new
    report_url=stdoutx
    PERMS = p+u+g+acl+selinux+xattrs
    CONTENTEX=sha512+ftype+p+u+g+n+acl+selinux+xattrs

    /hostroot/boot/        CONTENTEX
    /hostroot/root/\..* PERMS
    /hostroot/root/   CONTENTEX
    !/hostroot/root/\.kube
    !/hostroot/usr/src/
    !/hostroot/usr/tmp/

    /hostroot/usr/    CONTENTEX

    # OpenShift specific excludes
    !/hostroot/opt/
    !/hostroot/var
    !/hostroot/etc/NetworkManager/system-connections/
    !/hostroot/etc/mtab$
    !/hostroot/etc/.*~
    !/hostroot/etc/kubernetes/static-pod-resources
    !/hostroot/etc/kubernetes/test
    !/hostroot/etc/kubernetes/aide.*
    !/hostroot/etc/kubernetes/manifests
    !/hostroot/etc/kubernetes/kubelet-ca.crt
    !/hostroot/etc/docker/certs.d
    !/hostroot/etc/openvswitch/conf.db
    !/hostroot/etc/kubernetes/cni/net.d
    !/hostroot/etc/kubernetes/cni/net.d/*
    !/hostroot/etc/machine-config-daemon/currentconfig$
    !/hostroot/etc/machine-config-daemon/node-annotation.json*
    !/hostroot/etc/pki/ca-trust/extracted/java/cacerts$
    !/hostroot/etc/cvo/updatepayloads
    !/hostroot/etc/cni/multus/certs
    !/hostroot/etc/kubernetes/compliance-operator
    !/hostroot/etc/kubernetes/node-feature-discovery

    # Catch everything else in /etc
    /hostroot/etc/    CONTENTEX
  aide.conf: |-
    @@define DBDIR /hostroot/etc/kubernetes
    @@define LOGDIR /hostroot/etc/kubernetes
    database=file:@@{DBDIR}/aide.db.gz
    database_out=file:@@{DBDIR}/aide.db.gz.new
    gzip_dbout=yes
    verbose=6
    report_url=file:@@{LOGDIR}/aide.log.new
    report_url=stdoutx
    PERMS = p+u+g+acl+selinux+xattrs
    CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs

    /hostroot/boot/        CONTENT_EX
    /hostroot/root/\..* PERMS
    /hostroot/root/   CONTENT_EX
    !/hostroot/root/\.kube
    !/hostroot/usr/src/
    !/hostroot/usr/tmp/

    /hostroot/usr/    CONTENT_EX

    # OpenShift specific excludes
    !/hostroot/opt/
    !/hostroot/var
    !/hostroot/etc/NetworkManager/system-connections/
    !/hostroot/etc/mtab$
    !/hostroot/etc/.*~
    !/hostroot/etc/kubernetes/static-pod-resources
    !/hostroot/etc/kubernetes/test
    !/hostroot/etc/kubernetes/aide.*
    !/hostroot/etc/kubernetes/manifests
    !/hostroot/etc/kubernetes/kubelet-ca.crt
    !/hostroot/etc/docker/certs.d
    !/hostroot/etc/openvswitch/conf.db
    !/hostroot/etc/kubernetes/cni/net.d
    !/hostroot/etc/kubernetes/cni/net.d/*
    !/hostroot/etc/machine-config-daemon/currentconfig$
    !/hostroot/etc/machine-config-daemon/node-annotation.json*
    !/hostroot/etc/pki/ca-trust/extracted/java/cacerts$
    !/hostroot/etc/cvo/updatepayloads
    !/hostroot/etc/cni/multus/certs
    !/hostroot/etc/kubernetes/compliance-operator
    !/hostroot/etc/kubernetes/node-feature-discovery

    # Catch everything else in /etc
    /hostroot/etc/    CONTENT_EX
```

Vincent056:
@rhmdnd this should be ready for some reviews

Dockerfile.ci
```diff
@@ -1,27 +1,37 @@
...
 WORKDIR /go/src/github.com/openshift/file-integrity-operator
 
 ENV GOFLAGS="-mod=vendor"
 
+RUN dnf -y install git make gcc automake autoconf libtool flex gettext-devel e2fsprogs-devel audit-libs-devel libattr-devel flex bison zlib-devel libgcrypt-devel audit-libs-devel libacl-devel libselinux-devel libtool && dnf clean all
+
 COPY . .
 
+RUN git clone https://github.com/autoconf-archive/autoconf-archive.git && \
+    cp autoconf-archive/m4/*.m4 /usr/share/aclocal/
```
rhmdnd, line 13: Is this a dependency for building AIDE 0.18?

Makefile
```diff
@@ -161,6 +161,30 @@ else
 	$(eval OPENSHIFT_USER = $(shell oc whoami))
 endif
 
+
+.PHONY: aide-0.18
+aide-0.18:
+	rm -rf aide && \
+	git clone https://github.com/aide/aide.git && \
+	cd aide && \
```
rhmdnd, line 169: We have a `build` directory where we stash other related tools like `controller-gen` and `kustomize`. We could keep the entire repository in there so it gets cleaned up using `make clean`. Otherwise, we might need to go in and clean it up manually if we encounter a build failure.

Makefile
```diff
@@ -161,6 +161,30 @@ else
...
+	--disable-static \
+	--with-posix-acl \
+	--with-gcrypt \
+	--with-selinux \
+	--with-xattr \
+	--with-e2fsattrs \
+	--with-audit && \
+	$(MAKE) && \
+	cd .. && \
+	cp aide/aide ./build/bin/aide-0.18 && \
```
rhmdnd, line 183: +1 to putting this in the build directory

cmd/manager/daemon.go
```diff
@@ -67,17 +68,19 @@ type daemonConfig struct {
...
-	LogCollectorTimeout       int64
-	Interval                  int64
-	MaxBackups                int
-	LogCollectorCompress      bool
-	Local                     bool
-	Pprof                     bool
-	FileDir                   string
-	RunDir                    string
-	ConfigDir                 string
+
```
rhmdnd, line 71: nit: extra newline?

cmd/manager/loops.go
```diff
@@ -78,6 +78,52 @@ func aideLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errCha
...
+// The migrationCheckLoop checks if the migrated AIDE config file exists and if the new AIDE config works.
+func migrationCheckLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errChan chan<- error, wg *sync.WaitGroup) {
+	defer wg.Done()
+	migrateCtx, migrateCancel := context.WithCancel(ctx)
+	defer migrateCancel()
+
+	for {
+		select {
+		case <-migrateCtx.Done():
+			DBG("migrateLoop cancelled by the main routine!")
```
rhmdnd, line 90: This error will be written to logs, right? If so, we might want to think of a way to make the language less specific to the implementation details of the code. As a kubernetes administrator, how would I interpret this if I don't know the code? What do I need to do instead?

cmd/manager/loops.go
```diff
@@ -78,6 +78,52 @@ func aideLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errCha
...
+			return
+		default:
+			err := runAideMigrationCheckCmd(migrateCtx, conf)
+			aideResult := common.GetAideExitCode(err)
+			if aideResult == 0 {
+				LOG("AIDE migration check passed")
```
rhmdnd, line 96: We could explicitly say that the configuration we checked is safe to run using AIDE 0.18.0 (and that we'll just do that for users automatically or something).

cmd/manager/loops.go
```diff
@@ -78,6 +78,52 @@ func aideLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errCha
...
+				if err := reportOK(migrateCtx, conf, rt); err != nil {
+					// Considering this a non-fatal error right now.
+					LOG("failed reporting migration check result: %v", err)
+				}
+				return
+			} else if aideResult == 17 {
+				// This is an AIDE config line error.
+				logAndTryReportingDaemonError(migrateCtx, rt, conf, fmt.Sprintf("Detected error config during the migration check: %s",
```
rhmdnd, line 104: Similar comment as above, where we could say explicitly that we can't run this configuration with AIDE 0.18.0.

Does the migration check give us a line number that's causing the error? If so, we could give that to users in the log output so that know what they need to go change. That might be a stretch, but I'm just trying to think of ways to get users specific information that helps them work through the issue if we can't do it for them.

cmd/manager/loops.go
```diff
@@ -78,6 +78,52 @@ func aideLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errCha
...
+					common.GetAideErrorMessage(aideResult))+" %v", err)
+				if err := reportError(migrateCtx, fmt.Sprintf("Detected error config during the migration check: %s",
```
rhmdnd, line 106: nit: We could share the fmt.Sprintf between this line and 104 since it's the same.

cmd/manager/loops.go
```diff
@@ -78,6 +78,52 @@ func aideLoop(ctx context.Context, rt *daemonRuntime, conf *daemonConfig, errCha
...
+					common.GetAideErrorMessage(aideResult)), conf, rt); err != nil {
+					// Considering this a non-fatal error right now.
+					LOG("failed reporting migration check result: %v", err)
+				}
+				return
+			} else {
+				logAndTryReportingDaemonError(ctx, rt, conf, fmt.Sprintf("Error running migration check: %s",
+					common.GetAideErrorMessage(aideResult))+" %v", err)
+				if err := reportError(ctx, fmt.Sprintf("Error running migration check: %s",
```
rhmdnd, line 115: Similar comment as above on line 106.

tests/e2e/e2e_test.go
```diff
@@ -896,6 +896,51 @@ func TestFileIntegrityBadConfig(t *testing.T) {
...
+	// wait to go to active state. we should still be in active state but with an event.
+	err = waitForScanStatusWithTimeout(t, f, namespace, testName, v1alpha1.PhaseActive, retryInterval, timeout, 1)
+	if err != nil {
+		t.Errorf("Timeout waiting for scan status")
+	}
+
+	err = waitForAIDEConfigMigrationEvent(t, f, namespace, testName, "17")
+	if err != nil {
+		t.Errorf("Expected event to be found")
+	}
```
rhmdnd, line 940: Nice - thanks for adding this. So the remaining bit to test would be running AIDE 0.18.0, yeah? At this point, the test is assertion the configuration *should* work with 0.18.0, but we're still running 0.16.0.

rhmdnd (changes requested):
Took an initial pass, but I still need to come back and look through the config map controller and file integrity controller logic.

Posting the feedback I have for now.

Dockerfile.ci
Vincent056: yes, I think rhel might has rpm for this, but this builder does not

tests/e2e/e2e_test.go
Vincent056: correct

Vincent056:
> Took an initial pass, but I still need to come back and look through the config map controller and file integrity controller logic.
> 
> Posting the feedback I have for now.

thanks for the detail reviews!

build/Dockerfile
```diff
@@ -1,27 +1,38 @@
...
+RUN dnf -y install git make gcc automake autoconf libtool flex gettext-devel e2fsprogs-devel audit-libs-devel libattr-devel flex bison zlib-devel libgcrypt-devel audit-libs-devel libacl-devel libselinux-devel libtool && dnf clean all
+
 COPY . .
 
+RUN git clone https://github.com/autoconf-archive/autoconf-archive.git && \
+    cp autoconf-archive/m4/*.m4 /usr/share/aclocal/ 
+
 RUN make build
 
+RUN make aide-0.18
```
rhmdnd, line 18: Another alternative we talked about earlier was to build this using a separate ubi8 base image container and then copying it over, too.

Or, at least in this case, we could just copy it out of the fedora 40 image and see how that works.

build/Dockerfile
```diff
@@ -1,27 +1,38 @@
...
 ENV OPERATOR=/usr/local/bin/file-integrity-operator \
     USER_UID=1001 \
     USER_NAME=file-integrity-operator
 
+RUN mkdir -p /usr/local/libs
+
 # install operator binary
 COPY --from=builder /go/src/github.com/openshift/file-integrity-operator/build/bin/manager ${OPERATOR}
 COPY build/bin /usr/local/bin
+COPY --from=builder /go/src/github.com/openshift/file-integrity-operator/build/bin/aide-0.18 /usr/sbin/aide-0.18
```
rhmdnd, line 35: Then we don't need to pollute the builder container image with utilities and code unrelated to golang, or building FIO.

tests/e2e/helpers.go
rhmdnd, line 1: I think we'll also need to update the default AIDE configuration.

https://github.com/openshift/file-integrity-operator/blob/master/pkg/controller/fileintegrity/config_defaults.go#L15-L61

The way I understand the current logic is that FIO will continue to lay down AIDE configurations that are not compatible with 0.18.0. I think we'll want to change that as soon as possible, so long as the default AIDE configuration for 0.18.0 we want to use works on 0.16.0. That way we don't have the maintain the migration logic any longer than we absolutely have to.

tests/e2e/helpers.go
```diff
@@ -2173,7 +2273,6 @@ func logContainerOutput(t *testing.T, f *framework.Framework, namespace, name st
 		return
 	}
 
-	// Append operatorPods to pods
```
rhmdnd, line 2176: nit: unrelated change?

tests/e2e/helpers.go
```diff
@@ -1224,6 +1264,66 @@ func waitForFIStatusEvent(t *testing.T, f *framework.Framework, namespace, name,
 	return nil
 }
 
+func waitForFIMigrationErrAnnotation(t *testing.T, f *framework.Framework, namespace, name, expectedMessage string) error {
```
rhmdnd, line 1267: nit: we could use this as an assertion and not have to deal with checking the error output in the caller to fail the test, making it more of an assertion.

```go
func assertAIDEMigrationEmitsAnAlert(...)
```

And then a similar one below:

```go
func assertAIDEMigrationIsSuccessful(...)
```

But, I acknowledge this would be potentially different from existing conventions and we could do that in a separate PR if we decide to make that change across the entire functional test suite.

rhmdnd (changes requested):
Looking good - a few more comments inline now that I've had more time to walk through the controller code and the tests.

pkg/controller/fileintegrity/config.go
```diff
@@ -49,3 +53,151 @@ func prepareAideConf(in string) (string, error) {
...
+					if logLevel == "" && reportLevel == "" {
+						// if both are empty, it means the verbose value is invalid
+						return "", ignoredLines, fmt.Errorf("Invalid verbose value: %s", value)
+					}
+				} else {
+					newConfig.WriteString(newKey + "=" + value + "\n")
+				}
+			} else if _, ok := nonReplaceableKeys[key]; ok {
+				if !ignore {
+					return "", ignoredLines, fmt.Errorf("Deprecated option found: %s, please use %s separately", line, nonReplaceableKeys[key])
```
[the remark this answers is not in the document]
Vincent056: that's a good catch, the map should be ignore_list to all of those optios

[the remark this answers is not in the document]
Vincent056: https://github.com/aide/aide/blob/8ef6931719315a5f44cf6be11ccf73435822e2fe/doc/aide.conf.5#L360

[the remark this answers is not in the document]
Vincent056: in other word, ignore_list should replaced by
```
"report_ignore_added_attrs", "report_ignore_removed_attrs", "report_ignore_changed_attrs","report_ignore_e2fsattrs"
```

PR #539, opened 2024-07-08
implement rosa hcp e2e testing

rhmdnd:
- **Only set node selectors for OpenShift deployments**
- **[CMP-2624](https://issues.redhat.com//browse/CMP-2624): Implement support for running e2e tests on ROSA**

PR #540, opened 2024-07-09
Only set node selectors for OpenShift deployments

rhmdnd:
By default, the File Integrity Operator would set node selectors and
taints that would only allow it to schedule on `master` nodes. While
this is fine for OpenShift environments, some additional environments we
test on, like Red Hat OpenShift on AWS (ROSA), don't provide `master`
nodes at all.

To accomodate this in our testing and tooling, this commit removes the
node selector and taint from the default bundle, and moves it into an
OpenShift-specific bundle where it will still get used for OpenShift
deploy paths and bundles.

This will be more useful in a future patch that implements support for
running File Integrity Operator end-to-end testing on ROSA HCP, which
only provides `worker` nodes and not `master` nodes.


rhmdnd:
Debugging the e2e failures in https://github.com/openshift/file-integrity-operator/pull/541

rhmdnd:
This was already incorporated into https://github.com/openshift/file-integrity-operator/pull/539

PR #541, opened 2024-07-10
Exclude MCO registry secret from integrity checks

rhmdnd:
We're noticing some issues in CI where tests time out because node never
reach their desired state. We've also noticed MCO making changes to a
file that was previously monitored, causing failures we weren't
expecting.

This commit attempts to fix the issue by excluding the MCO file in
question so it stops raising false negative results in the e2e suite.


Vincent056:
> We're noticing some issues in CI where tests time out because node never reach their desired state. We've also noticed MCO making changes to a file that was previously monitored, causing failures we weren't expecting.
> 
> This commit attempts to fix the issue by excluding the MCO file in question so it stops raising false negative results in the e2e suite.

https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1720621801639479 we should be good to exclude that

PR #543, opened 2024-07-15
Bump up dependency

Vincent056:
Let's bump up dependency in one place to fix CI


Vincent056:
https://github.com/openshift/release/pull/54443 has to merge to fix some of CI failures

cmd/manager/operator.go
```diff
@@ -366,22 +385,42 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
 		log.Info("Install prometheus-operator in your cluster to create ServiceMonitor objects")
 		return nil
 	}
+	serviceAccountName := "file-integrity-operator"
 
+	// Get the name of the secret associated with the service account
+	secretName, err := getSecretNameForServiceAccount(kubeClient, namespace, serviceAccountName)
+	if err != nil {
+		if kerr.IsNotFound(err) {
+			fmt.Println("Secret not found - restarting, as the service may have just been created")
```
rhmdnd, line 394: We should try using logging for these instead of printing to stdout for consistency.

cmd/manager/operator.go
```diff
@@ -391,12 +391,12 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
 	secretName, err := getSecretNameForServiceAccount(kubeClient, namespace, serviceAccountName)
 	if err != nil {
 		if kerr.IsNotFound(err) {
-			fmt.Println("Secret not found - restarting, as the service may have just been created")
+			log.Error(err, "Secret not found - restarting, as the service may have just been created")
 		} else {
-			fmt.Println("Error retrieving secret:", err)
+			log.Error(err, "Error retrieving secret")
```
rhmdnd, line 396: Will this just keep looping like the case above?

Vincent056: that's a good point, I think the error should actually be returned, because we would want to retry here, didn't catch this while reviewing https://github.com/ComplianceAsCode/compliance-operator/pull/491/files#diff-51bd220db414457ea4470ea5d71465bd07e1c789d156c0104e1fce2f3ddbcb1aR771

config/rbac/metrics_secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: metrics-token
+  namespace: security-profiles-operator
```
rhmdnd, line 5: This should be updated to use `file-integrity-operator`, I think.

cmd/manager/operator.go
```diff
@@ -367,21 +367,29 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
...
+	for i := range serviceMonitor.Spec.Endpoints {
 		if serviceMonitor.Spec.Endpoints[i].Port == metrics.ControllerMetricsServiceName {
 			serviceMonitor.Spec.Endpoints[i].Path = metrics.HandlerPath
 			serviceMonitor.Spec.Endpoints[i].Scheme = "https"
-			serviceMonitor.Spec.Endpoints[i].BearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+			serviceMonitor.Spec.Endpoints[i].Authorization = &monitoring.SafeAuthorization{
+				Type: "Bearer",
+				Credentials: &v1.SecretKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: "metrics-token",
```
rhmdnd, line 380: Is this token for metrics, or monitoring?

rhmdnd (changes requested):
A few suggestions inline. Thanks!

config/rbac/metrics_secret.yaml
Vincent056: oops let me have that fixed

cmd/manager/operator.go
Vincent056: It is token for metric service account which allowed to view metrics endpoints

cmd/manager/operator.go
```diff
@@ -390,13 +390,8 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
...
 	secretName, err := getSecretNameForServiceAccount(kubeClient, namespace, serviceAccountName)
 	if err != nil {
-		if kerr.IsNotFound(err) {
-			fmt.Println("Secret not found - restarting, as the service may have just been created")
-		} else {
-			fmt.Println("Error retrieving secret:", err)
-		}
-	} else {
-		fmt.Println("Secret name:", secretName)
+		log.Error(err, "Error getting secret name for service account")
```
rhmdnd: Another thing we can incorporate here is the name of the secret and service account. Thinking about the information that an administrator would need to fix the issue just by reading the log. Otherwise they'll either need to guess which secret we're implying here or read the code.

I'm ok doing that in a follow up through before the next release.

rhmdnd: Oh - wait. I see we're removing this entirely in the next commit.

cmd/manager/operator.go
```diff
@@ -385,16 +366,9 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
...
 
-	// Get the name of the secret associated with the service account
-	secretName, err := getSecretNameForServiceAccount(kubeClient, namespace, serviceAccountName)
-	if err != nil {
-		log.Error(err, "Error getting secret name for service account")
-		return err
-	}
+	serverName := fmt.Sprintf("metrics.%s.svc", namespace)
 	serviceMonitor := common.GenerateServiceMonitor(service)
-	serverNameStr := "metrics." + namespace + ".svc"
```
rhmdnd: One thing I try to be mindful of is reducing the amount of non-functional change lumped into a broader patch, since it can muddy the git log and make it harder to use git has a tool for dissecting historical changes.

However, we can and should make non-functional improvements for readability, maintenance, consistency, etc... but I'm inclined to make those changes in a dedicated commit with clear reasoning for the change outlined in the commit message.

rhmdnd (approved):
/lgtm

Thank you for working through this mess and getting CI working with updated deps!

rhmdnd:
Applying additional labels so we can get these dependencies updated. Note that we're going to add some additional functional testing for monitoring in a subsequent patch.

PR #550, opened 2024-07-29
AIDE 0.18 e2e

Vincent056:
This PR adds the dockerfile that builds operator image with the AIDE 0.18, also replacing 0.16 config file with the 0.18 conf file using the migration logic depends on value of AIDE_VERSION. We will use 0.16 as default value if not set. This is for us to have CI to test AIDE config.

We will need to merge this one and test on https://github.com/openshift/release/pull/54911

### DIFF
--- a/pkg/controller/fileintegrity/config_defaults.go
+++ b/pkg/controller/fileintegrity/config_defaults.go
@@ -1,5 +1,7 @@
 package fileintegrity
 
+import "os"
+
 const aideLogPath = "/hostroot/etc/kubernetes/aide.log"
 
 var aideReinitContainerScript = `#!/bin/sh
@@ -12,8 +14,30 @@ var aidePauseContainerScript = `#!/bin/sh
 	wait $PID || true
 `
 
-// NOTE: Needs to be in sync with `testAideConfig` in test/e2e/helpers.go, except for the heading comment.
-var DefaultAideConfig = `@@define DBDIR /hostroot/etc/kubernetes
+var DefaultAideConfigCommonStart018 = `@@define DBDIR /hostroot/etc/kubernetes
+@@define LOGDIR /hostroot/etc/kubernetes
+database_in=file:@@{DBDIR}/aide.db.gz
+database_out=file:@@{DBDIR}/aide.db.gz.new
+gzip_dbout=yes
+log_level=warning
+report_level=changed_attributes
+report_url=file:@@{LOGDIR}/aide.log.new
+report_url=stdout
+PERMS = p+u+g+acl+selinux+xattrs
+CONTENTEX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
+
+/hostroot/boot/        CONTENTEX
+/hostroot/root/\..* PERMS
+/hostroot/root/   CONTENTEX
+!/hostroot/root/\.kube
+!/hostroot/usr/src/
+!/hostroot/usr/tmp/
+`
+
+var DefaultAideConfigCommonEnd018 = `# Catch everything else in /etc
+/hostroot/etc/    CONTENTEX`
+
+var DefaultAideConfigCommonStart = `@@define DBDIR /hostroot/etc/kubernetes
 @@define LOGDIR /hostroot/etc/kubernetes
 database=file:@@{DBDIR}/aide.db.gz
 database_out=file:@@{DBDIR}/aide.db.gz.new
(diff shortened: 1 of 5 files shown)


pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -56,6 +85,11 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/kubernetes/compliance-operator
 !/hostroot/etc/kubernetes/node-feature-discovery
 !/hostroot/etc/mco/internal-registry-pull-secret.json$
+`
 
-# Catch everything else in /etc
-/hostroot/etc/    CONTENT_EX`
+func GetAideConfigDefault() string {
+	if os.Getenv("AIDE_VERSION") == "0.18" {
```
rhmdnd, line 91: 🤔 I wonder if we can get this information from the RPMs available on the host, instead of setting an env?

Vincent056: this would be hard to do, as we need to run a pod with a command to fetch it, current step is to use a env in the aide-0.18 dockerfile to set that

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -12,8 +14,30 @@ var aidePauseContainerScript = `#!/bin/sh
...
 `
 
-// NOTE: Needs to be in sync with `testAideConfig` in test/e2e/helpers.go, except for the heading comment.
-var DefaultAideConfig = `@@define DBDIR /hostroot/etc/kubernetes
+var DefaultAideConfigCommonStart018 = `@@define DBDIR /hostroot/etc/kubernetes
+@@define LOGDIR /hostroot/etc/kubernetes
+database_in=file:@@{DBDIR}/aide.db.gz
+database_out=file:@@{DBDIR}/aide.db.gz.new
+gzip_dbout=yes
+log_level=warning
```
rhmdnd, line 22: Can everything under here go into `DefaultAideConfigExclude`? The only two differences I see are `database_in` and `verbose`?

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -12,8 +14,30 @@ var aidePauseContainerScript = `#!/bin/sh
...
+report_level=changed_attributes
+report_url=file:@@{LOGDIR}/aide.log.new
+report_url=stdout
+PERMS = p+u+g+acl+selinux+xattrs
+CONTENTEX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
+
+/hostroot/boot/        CONTENTEX
+/hostroot/root/\..* PERMS
```
rhmdnd, line 30: Looks like this is the same for 0.16.0 and 0.18.0.

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -12,8 +14,30 @@ var aidePauseContainerScript = `#!/bin/sh
...
+/hostroot/root/   CONTENTEX
+!/hostroot/root/\.kube
```
rhmdnd, line 32: Same with these three lines.

pkg/controller/fileintegrity/config_defaults.go
```diff
@@ -56,6 +85,11 @@ CONTENT_EX = sha512+ftype+p+u+g+n+acl+selinux+xattrs
 !/hostroot/etc/kubernetes/compliance-operator
 !/hostroot/etc/kubernetes/node-feature-discovery
 !/hostroot/etc/mco/internal-registry-pull-secret.json$
+`
 
-# Catch everything else in /etc
-/hostroot/etc/    CONTENT_EX`
+func GetAideConfigDefault() string {
+	if os.Getenv("AIDE_VERSION") == "0.18" {
+		return DefaultAideConfigCommonStart018 + DefaultAideConfigExclude + DefaultAideConfigCommonEnd018
```
rhmdnd, line 92: Random question - does the ordering of the config matter? Do we have to chunk it into beginning, middle, and end? If not, we could have a version-specific chunk, and a common chunk instead of sandwiching the common bits in between two version specific variables.

pkg/controller/fileintegrity/config_defaults.go
rhmdnd: Ok - thanks for the clarification.

pkg/controller/fileintegrity/config_defaults.go
Vincent056: `CONTENT_EX` is being replaced with `CONTENTEX` and `log_level` are added along with `report_level`

pkg/controller/fileintegrity/config_defaults.go
Vincent056: yes, I think it might affect what sequence are file being scanned and processed, giving what being scanned/excluded according to top down sequence.

Vincent056: we want to scan system directory before user directory for example `/hostroot/etc/`  before `/hostroot/usr/`

rhmdnd:
Ok - looks good. I only have one minor suggestion on the new e2e suite for AIDE 0.18.0 to make it optional and then we can start exercising it against this PR.

rhmdnd (approved):
/lgtm

This is passing tests with 0.16.0 and gives us a start for all the 0.18.0 migration work.

Approving so we can get this landed and then land the CI change to incorporate an AIDE 0.18.0 job.

https://github.com/openshift/release/pull/54911

rhmdnd:
No docs or px needed yet since this is laying down the test plumbing. No functional changes.

/px-approved
/docs-approved

PR #556, opened 2024-08-13
Fix a missing back slash in aide0.18 dockerfile

Vincent056:
We are adding a missing `\` in the aide 0.18 dockerfile.


rhmdnd (approved):
/lgtm

PR #577, opened 2024-10-15
Use latest in the CSV instead of a version tag

rhmdnd:
This makes it so that the replace logic also replaces the deployment
image in addition to the related image. This is necessary CI to work
properly, because it needs to be able to replace all images in the CSV
with a build from CI.


yuumasato:
This seems to work, but there were a lot of test failures, :cry:

yuumasato:
Yes, I think we need this change.

But we also need this change in the FIO release PR too.

yuumasato:
I have cherry-picked this commit to the release PR.

yuumasato (approved):
/approve

PR #584, opened 2024-10-18
OCPBUGS-42807: Preserve metrics when pod crashes

Vincent056:
The metric `file_integrity_operator_node_failed` was being reset after the File Integrity Operator pod restarted, causing the `NodeHasIntegrityFailure` alert to clear despite failures persisting in `fileintegritynodestatuses`. This fix ensures that the operator retrieves the status of existing node failures and re-triggers the alert as needed, maintaining consistency in node failure detection.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	dto "github.com/prometheus/client_model/go"
 
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -59,6 +60,10 @@ const (
 	HandlerPath                        = "/metrics-fio"
 	ControllerMetricsServiceName       = "metrics-fio"
 	ControllerMetricsPort        int32 = 8585
+)
+
+var (
+	ErrNoMetrics = errors.New("metric value is nil")
 )
 
 // Metrics is the main structure of this package.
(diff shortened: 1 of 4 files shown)


tests/e2e/helpers.go
```diff
@@ -287,6 +287,46 @@ spec:
...
+	}
+	pod := podList.Items[0]
+	t.Logf("Killing operator pod %s", pod.Name)
+	err = f.KubeClient.CoreV1().Pods(namespace).Delete(goctx.TODO(), pod.Name, metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+
+	// wait for the pod to be recreated
+	err = wait.Poll(pollInterval, pollTimeout, func() (bool, error) {
```
yuumasato, line 308: @Vincent056 We may need to increase the `pollTimeout`.

xiaojiey:
Pre-merge verification pass with https://github.com/openshift/file-integrity-operator/pull/584
steps:
1. Install File Integrity Operator and check "Enable Operator Recommended Monitoring" checkbox.
2. Create FileIntegrity custom resource and wait for the pods to initialize fine.
3. Simulate a failure condition and wait for NodeHasIntegrityFailure alert to stream.
4. Delete the file-integrity-operator pod.
5. After couple of minutes the alert NodeHasIntegrityFailure still exists and the "oc get fileintegritynodestatuses" reports failure for the node on which failure condition was simulated.

rhmdnd:
Looks like we got a failure with the intial delay feature:

```
e2e_test.go:207: Testing that we don't get a ready daemonSet before the initialDelaySeconds
e2e_test.go:210: We should expect a timed out waiting for daemonSet aide-e2e-test-initialdelay 
```

pkg/controller/metrics/metrics.go
```diff
@@ -303,6 +308,33 @@ func (m *Metrics) IncFileIntegrityNodeStatus(condition, node string) {
...
+		return 0, false, fmt.Errorf("failed to get metric for node: %s", node)
+	}
+	c := make(chan prometheus.Metric, 1)
+	ctr.Collect(c)
+	// Ensure the channel is not empty before reading
+	select {
+	case metric := <-c:
+		dto := dto.Metric{}
+		if err := metric.Write(&dto); err != nil {
+			return 0, false, fmt.Errorf("failed to write metric: %w", err)
```
rhmdnd, line 326: At first glance, I assumed this method only retrieved metrics based on the name, but it's actually performing a write, which seems more like an update.

Vincent056: no, it is not a write, just need to write out to the channel so we can see the output

PR #590, opened 2024-10-21
Increase e2e timeout for checking node success

rhmdnd:
We're seeing several tests fail with the following:

  ERROR: nodes weren't in good state: status.nodeStatus for node ip-10-0-30-92.us-west-2.compute.internal NOT SUCCESS: But instead Failed

This commit increases the timeout from 5 and 10 minutes to 15 minutes so
that the nodes have more time to report back successfully.


rhmdnd:
/test e2e-aws

Retesting to see if we can get a second successful run.

rhmdnd:
/test e2e-aws

Most recent failure was unrelated to this change (cluster provisioning).

yuumasato (approved):
/approve

PR #594, opened 2024-10-30
Annotate the file to nudge after operator builds

yuumasato:
Whenever an operator build on push events is successful the PULLSPEC reference in update_csv.py will be updated.


rhmdnd (approved):
/lgtm

PR #607, opened 2024-11-21
OCPBUGS-36483: Fix new added node issue

Vincent056:
`/etc/kubernetes/static-pod-resource` is being created after a new node is added, this cause the link count change in the aide checks and resulting a failures. The PR added a init pod to check for that before running aide daemonset.

### DIFF
--- a/pkg/controller/fileintegrity/fileintegrity_controller.go
+++ b/pkg/controller/fileintegrity/fileintegrity_controller.go
@@ -869,8 +869,7 @@ func aideDaemonset(dsName string, fi *v1alpha1.FileIntegrity, operatorImage stri
 								fi`,
 							},
 							SecurityContext: &corev1.SecurityContext{
-								Privileged: &priv,
-								RunAsUser:  &runAs,
+								RunAsUser: &runAs,
 							},
 							Resources: corev1.ResourceRequirements{
 								Requests: corev1.ResourceList{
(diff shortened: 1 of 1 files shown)


xiaojiey:
/label qe-approved
The fileintegritynodestatus will be "Succeeded" for newly added nodes

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -856,6 +856,41 @@ func aideDaemonset(dsName string, fi *v1alpha1.FileIntegrity, operatorImage stri
...
+							Command: []string{
+								"sh",
+								"-c",
+								`if [ ! -d "/host/etc/kubernetes/static-pod-resources" ]; then
+									echo "Directory /etc/kubernetes/static-pod-resources does not exist, exiting.";
+									exit 1;
+								fi`,
+							},
+							SecurityContext: &corev1.SecurityContext{
+								Privileged: &priv,
```
rhmdnd, line 872: Does this need to be privileged? Or can we reduce the permissions required to run this container?

rhmdnd (approved):
Thoughts on how we can protect this from regressions in CI?

Vincent056:
> Thoughts on how we can protect this from regressions in CI?

I think this fixes the instability of the scaling e2e test @rhmdnd

PR #649, opened 2025-05-01
CMP-3381: Update the bundle pipeline to only run for specific changes

rhmdnd:
We only really need to build the bundle image when the operator image
reference changes, or if the build pipline configuration change for the
bundle build.


Vincent056 (approved):
/lgtm

PR #650, opened 2025-05-01
CMP-3381: Configure Konflux to bump image SHAs

rhmdnd:
https://github.com/openshift/file-integrity-operator/pull/598 refactored
the update_csv.py to update_csv.go. This commit updates the nudge to
account for that so we can have Konflux propose image updates
automatically.


Vincent056 (approved):
/lgtm

PR #660, opened 2025-05-16
OCPBUGS-56409: Fix FIPS for aide container

Vincent056:
This update introduces an LD_PRELOAD-based guard library to block MD5 usage by AIDE in FIPS mode, enforcing strict compliance independently of AIDE’s internal error handling. The Makefile and build scripts now build and package this guard, injecting it during AIDE database initialization. Enhanced error handling clearly identifies guard-triggered terminations as FIPS compliance errors.

This ensures robust FIPS enforcement, eliminating accidental MD5 use by AIDE.

Background:
• [Red Hat FIPS Compliance](https://access.redhat.com/compliance/fips?extIdCarryOver=true&sc_cid=701f2000001Css5AAC)

From libgcrypt-1.10.0-10 onward, MD5 and other non-FIPS algorithms no longer fail automatically. A new FIPS service indicator API (GCRYCTL_FIPS_SERVICE_INDICATOR_MD) classifies algorithms as approved (GPG_ERR_NO_ERROR) or merely allowed (GPG_ERR_NOT_SUPPORTED). MD5 is classified as merely allowed.

### DIFF
--- a/build/guard/libaide_md5_guard.c
+++ b/build/guard/libaide_md5_guard.c
@@ -1,0 +1,148 @@
+#define _GNU_SOURCE
+#include <dlfcn.h>
+#include <gcrypt.h>
+#include <gpg-error.h>
+#include <pthread.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+/* We intercept gcry_md_open and gcry_md_enable */
+static gcry_error_t (*real_md_open)(gcry_md_hd_t *, int, unsigned int);
+static gcry_error_t (*real_md_enable)(gcry_md_hd_t, int);
+
+/* One-time sync for resolving symbols */
+static pthread_once_t once = PTHREAD_ONCE_INIT;
+
+/*
+ * Small async-safe logger that writes directly to stderr,
+ * avoiding recursion if we're hooking I/O calls.
+ */
+__attribute__((format(printf, 1, 2)))
+static void logln(const char *fmt, ...)
+{
+    char buf[256];
+    va_list ap;
+    va_start(ap, fmt);
+    int n = vsnprintf(buf, sizeof(buf), fmt, ap);
+    va_end(ap);
+    if (n > (int)sizeof(buf) - 1) {
+        n = (int)sizeof(buf) - 1;
+    }
+    buf[n++] = '\n';
+    write(STDERR_FILENO, buf, n);
+}
+
+/*
+ * We only resolve the real gcry_md_* symbols once.
+ * No references to gcry_fips_mode_active here, to avoid symbol issues.
+ */
+static void resolve_symbols(void)
+{
+    real_md_open   = dlsym(RTLD_NEXT, "gcry_md_open");
+    real_md_enable = dlsym(RTLD_NEXT, "gcry_md_enable");
+}
+
+/*
+ * We check the kernel's FIPS flag: /proc/sys/crypto/fips_enabled
+ * If it's '1', we treat the system as in FIPS mode.
+ */
+static int in_fips_mode(void)
+{
+    FILE *fp = fopen("/proc/sys/crypto/fips_enabled", "r");
+    if (!fp) {
+        // If missing, assume not in FIPS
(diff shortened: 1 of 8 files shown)


build/guard/libaide_md5_guard.c
```diff
@@ -0,0 +1,148 @@
...
+ */
+static int in_fips_mode(void)
+{
+    FILE *fp = fopen("/proc/sys/crypto/fips_enabled", "r");
+    if (!fp) {
+        // If missing, assume not in FIPS
+        return 0;
+    }
+    char c = '0';
+    if (fread(&c, 1, 1, fp) != 1) {
```
rhmdnd, line 59: Not for other reviewers since it took me a minute to figure this out.

This is pretty defensive in that we're only reading a single byte and then confirming we're only getting a single byte back. If that's not the case, assume we're not in FIPS mode.

build/guard/libaide_md5_guard.c
```diff
@@ -0,0 +1,148 @@
...
+
+    /* In FIPS mode => allow empty multi-hash context (algo=0) */
+    if (algo == 0) {
+        return real_md_open(hd, 0, flags);
+    }
+
+    /* If a single-hash context of MD5 is requested, block. */
+    if (algo == GCRY_MD_MD5) {
+        const char *name = "MD5";
+        if (getenv("AIDE_GUARD_SOFT")) {
```
rhmdnd, line 98: Should we use `secure_getenv()` here instead?

https://man7.org/linux/man-pages/man3/getenv.3.html

build/guard/libaide_md5_guard.c
```diff
@@ -0,0 +1,148 @@
...
+    return (c == '1');
+}
+
+/*
+ * Hook for gcry_md_open:
+ * - If not in FIPS mode, pass everything through.
+ * - If in FIPS and algo=MD5, block.
+ * - Otherwise (SHA256, SHA512, etc.), allow.
+ */
+gcry_error_t gcry_md_open(gcry_md_hd_t *hd, int algo, unsigned int flags)
```
rhmdnd, line 73: Looks like there are other entry points related to MD5 (`gcry_md_read` and `gcry_md_write`).

Do we need to cover either of those, or is that irrelevant if we assume these are called first?

build/guard/libaide_md5_guard.c
```diff
@@ -0,0 +1,148 @@
...
+    /* Outside FIPS => no block */
+    if (!in_fips_mode()) {
+        return real_md_enable(hd, algo);
+    }
+
+    /* In FIPS => block MD5 */
+    if (algo == GCRY_MD_MD5) {
+        const char *name = "MD5";
+        if (getenv("AIDE_GUARD_SOFT")) {
+            logln("[md-guard] Attempt to enable %s in FIPS (pid %d) — soft block",
```
rhmdnd, line 137: The process ID here will be logged. Do we need to take any additional security measures to protect that information?

build/guard/libaide_md5_guard.c
Vincent056: https://github.com/aide/aide/blob/2677eff4c7126df76e98148023cdbdc2b18f278f/src/md.c#L102
it looks like those two will be used first in the AIDE in the init function

build/guard/libaide_md5_guard.c
```diff
@@ -0,0 +1,148 @@
...
+    if (algo == GCRY_MD_MD5) {
+        const char *name = "MD5";
+        if (getenv("AIDE_GUARD_SOFT")) {
+            logln("[md-guard] Attempt to open %s in FIPS (pid %d) — soft block",
+                  name, getpid());
+            return GPG_ERR_NOT_SUPPORTED;
+        }
+        logln("[md-guard] Attempt to open %s in FIPS (pid %d) — terminating",
+              name, getpid());
+        _exit(64);
```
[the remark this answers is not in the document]
Vincent056: if atexit is not being used in aide, maybe we could keep using it?

rhmdnd (approved):
/lgtm

PR #670, opened 2025-05-26
Rename branch and components for release-1-3

yuumasato:
After copying the pipelines from master branch and setting up the application and components let's rename branch and references to match them.


.tekton/file-integrity-operator-1-3-pull-request.yaml
```diff
@@ -8,13 +8,13 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "release-1.3"
+      == "master"
```
rhmdnd, line 11: I think we want these using `release-1.3` so that the pipelines run on the `release-1.3` branch. Otherwise, the FIO 1.3 application is going to be building from `master`, right?

PR #673, opened 2025-05-27
Fix bundle file path for cel expression

rhmdnd:
We updated this pipeline to use the release-1.3 branch, but we were using the older master pipeline name for the trigger. Instead, we want to make sure we trigger this pipeline when it's changed, so let's update the file path to point to the correct pipeline.


Vincent056 (approved):
/lgtm

PR #674, opened 2025-05-28
Update digest for operator image

rhmdnd:
Since we now have a separate pipeline for the 1.3 releases, we need to update the image digest to use the image built from that pipeline in the bundle image. Otherwise, we're going to be using the operator image built from master.


rhmdnd (commented):
/lgtm

PR #675, opened 2025-05-28
Update pipeline runs to include branch name as image tag

rhmdnd:
By adding the branch name to the most recent builds on push events
through Konflux, we can make it easier to fetch the latest builds on a
given branch. For example, we can fetch them using the `master` tag or
the latest 1.3 release candidate using `release-1.3` for both the bundle
and the operator.

Adding this tag makes it easier to fetch the latest images from other
locations, like the FBC repository where we need to find the digest to
build the appropriate index image.


Vincent056 (approved):
/lgtm

PR #677, opened 2025-05-28
Adds polling step to wait for operator image tag before bundle build

Vincent056:
Introduces a polling mechanism to ensure the required operator image tag is available on Quay before proceeding with bundle build

### DIFF
--- a/.tekton/file-integrity-operator-bundle-pull-request.yaml
+++ b/.tekton/file-integrity-operator-bundle-pull-request.yaml
@@ -165,6 +165,52 @@ spec:
       workspaces:
       - name: basic-auth
         workspace: git-auth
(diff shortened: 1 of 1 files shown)


yuumasato:
@Vincent056 Based on https://konflux.pages.redhat.com/docs/users/building/redundant-rebuilds.html

I think we could change the annotation on line 10 `pipelinesascode.tekton.dev/on-cel-expression` to only trigger the pipeline when `bundle-hack/update_csv.go` is changed:

```yaml
    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
      == "master" && "bundle-hack/update_csv.go".pathChanged()
```

What do you think?

PR #687, opened 2025-06-05
Rename the 1-3 pipelines to release-1-3

yuumasato:
The application and components have release-1-3 suffix.
Additionally, add the specific ServiceAccount that needs to be used in each pipeline.

This is being done as part of migration to new Konflux cluster.


rhmdnd (approved):
Konflux changes are going to fail until this is merged. 

Thanks for the PR.

PR #703, opened 2025-06-11
CMP-3424: Remove dependabot configuration

rhmdnd:
Konflux is configured to build FIO on each PR and merge, but it also
managed dependencies:

  https://konflux-ci.dev/docs/mintmaker/user/

Instead of having two different tools step on each other and propose
multiple PRs for the same updates, let's just pick one to manage
dependencies.


rhmdnd:
@Vincent056 @yuumasato this should help reduce the review queue for dependencies.

PR #739, opened 2025-06-30
Make bundle pipelines runs more restricted

yuumasato:
Only run bundle pipelines when there are changes to bundle relevant files.


Vincent056 (approved):
/lgtm

PR #746, opened 2025-07-08
Remove unnecessary Pipeline configuration for Konflux

rhmdnd:
Initially, we used these pipeline configuration files for operator and
bundle builds. Now, we're using the -dev suffix in the pipeline
application and components to make it easier to delineate dev builds
from release builds.

This commit removes the original pipeline configuration files so we can
simplify the amount of information reported on each PR, especially since
we should only be building for the -dev pipeline anyway.


yuumasato (approved):
/lgtm

Vincent056 (approved):
/lgtm

PR #753, opened 2025-07-23
Fix blank version label in bundle image

yuumasato:
The ARG variables are reset at the end of the scope they were defined.
This ensures version ARGs are available through out the build.

### DIFF
--- a/bundle.openshift.Dockerfile
+++ b/bundle.openshift.Dockerfile
@@ -1,15 +1,22 @@
+ARG FIO_OLD_VERSION="1.3.5"
+ARG FIO_NEW_VERSION="1.3.6"
+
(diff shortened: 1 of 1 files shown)


rhmdnd (approved):

rhmdnd:
We'll need to backport this to the `release-1.3` branch, too?

yuumasato:
> We'll need to backport this to the `release-1.3` branch, too?

Hmm, we can, so that any z-stream release has correct version label.
But in context of FIO 1.3.6 it is not required.

rhmdnd (approved):
/lgtm

Consistent with what we're doing in CO: 

https://github.com/ComplianceAsCode/compliance-operator/blob/master/bundle.openshift.Dockerfile

PR #787, opened 2025-11-19
CMP-3954: Add redhat.repo file

rhmdnd:
This will make it so Konflux keeps the RPM lock file up to date,
reducing the likelihood that we'll get stale RPMs when we do builds
(reducing manualy work).


rhmdnd:
@yuumasato should be ready for a review (we'll need to back port this to release-1.3, too I think so we get updated RPMs there)

Vincent056 (approved):
/lgtm

build/rpms.in.yaml
Vincent056, line 1: we also need libgcrypt-devel

Vincent056, line 1: do we need to remove arm, I was checking 1.3.5 manifest found

```
be670261379ff448f563
vincent@fedora:~/TEST$ cat manifests/file-integrity-operator.v1.3.5.clusterserviceversion.yaml | grep arch
    operatorframework.io/arch.amd64: supported
    operatorframework.io/arch.ppc64le: supported
    operatorframework.io/arch.s390x: supported
```

yuumasato: Yes, remove `aarch64`:
https://access.redhat.com/errata/RHBA-2025:11535

yuumasato: Are you generating the lock files with `--bare`?
I'm not sure what the MintMaker invocation is, but I don't think we need `libcgrypt` or `libgcrypt-devel` listed as an explicit package to be listed and installed.

When I run it without `--bare` `libgcrypt` isn't even listed:
[rpms.lock.yaml](https://github.com/user-attachments/files/23654236/rpms.lock.yaml)

yuumasato (changes requested):
FIO already has 2 `rpms.lock.yaml` one for the operator image and another for the bundle:
- https://github.com/openshift/file-integrity-operator/blob/master/bundle-hack/rpms.lock.yaml
- https://github.com/openshift/file-integrity-operator/blob/master/konflux/rpms.lock.yaml

We can consolidate both to use the same repo file, we would need to add an `rpms.in.yaml` for each one, and point to the single repo file.

yuumasato:
> FIO already has 2 `rpms.lock.yaml` one for the operator image and another for the bundle:
> 
>     * https://github.com/openshift/file-integrity-operator/blob/master/bundle-hack/rpms.lock.yaml
> 
>     * https://github.com/openshift/file-integrity-operator/blob/master/konflux/rpms.lock.yaml
> 
Actually, the Bundle image doesn't install RPMs, so it doesn't need a lock file.

The lock file in use is the one in the `konflux` directory:
https://github.com/openshift/file-integrity-operator/blob/b526d4c801e49ab5ddd79afcaa5c970827571f42/.tekton/file-integrity-operator-dev-pull-request.yaml#L41

build/rpms.in.yaml
Vincent056: do we need `libgcrypt-devel` for https://github.com/openshift/file-integrity-operator/blob/b526d4c801e49ab5ddd79afcaa5c970827571f42/build/Dockerfile.openshift#L8C20-L8C35

yuumasato, line 1: Based on the PRs created by MintMaker the`rpms.lock.yaml` is created in the same directory as `rpms.in.yaml`. 
So you can either move this file to `konflux` directory, or change the pipeline to use the lock file from the `build` directory.
https://github.com/openshift/file-integrity-operator/blob/09be206e10c8f8e95c56601cde7830b3594a1035/.tekton/file-integrity-operator-dev-pull-request.yaml#L41

yuumasato:
Superseded by https://github.com/openshift/file-integrity-operator/pull/926

PR #790, opened 2025-11-20
Update rpms.lock.yaml to latest

Vincent056:


Vincent056:
updating in https://github.com/openshift/file-integrity-operator/pull/778 instead

PR #792, opened 2025-11-20
Add PR bundle image

Vincent056:
This commit introduces a new task to the Tekton pipeline that waits for the operator image to become available. The task includes parameters for the operator image, maximum wait time, and check interval, along with a script to handle the waiting logic. The init task is updated to run after this new task, ensuring proper sequencing in the pipeline.

### DIFF
--- a/.tekton/file-integrity-operator-bundle-dev-pull-request.yaml
+++ b/.tekton/file-integrity-operator-bundle-dev-pull-request.yaml
@@ -135,6 +135,25 @@ spec:
       name: CHAINS-GIT_COMMIT
       value: $(tasks.clone-repository.results.commit)
     tasks:
+    - name: wait-for-operator-image
+      params:
+      - name: OPERATOR_IMAGE
+        value: quay.io/redhat-user-workloads/ocp-isc-tenant/file-integrity-operator-dev:on-pr-$(params.revision)
+      - name: ARCHITECTURES
+        value: "linux/amd64,linux/ppc64le,linux/s390x"
+      - name: MAX_WAIT_TIME
+        value: "30m"
+      - name: CHECK_INTERVAL
+        value: "30s"
+      taskRef:
+        resolver: git
+        params:
+        - name: url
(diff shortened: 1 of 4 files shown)


yuumasato:
This is very interesting @Vincent056
The EC tests didn't flag `wait-for-operator-image` as untrusted, I wonder if it will continue that way if merged.

.tekton/file-integrity-operator-bundle-dev-pull-request.yaml
```diff
@@ -127,6 +127,25 @@ spec:
       name: CHAINS-GIT_COMMIT
       value: $(tasks.clone-repository.results.commit)
     tasks:
+    - name: wait-for-operator-image
```
yuumasato, line 138: @Vincent056 I'm starting to think that we need to have a separate pipeline that builds the Bundle image from the PR changes.

I say this because ideally we should be building the images only when relevant changes to each image is made:
- We only build the operator image when changes for the operator code are made
- We should only build the bundle image when relevant changes for the bundle are made.

There will be cases when we may only need or want to build the bundle image. In this case there is no operator image build to wait for.

yuumasato:
@Vincent056 I have tried a different but similar approach in https://github.com/openshift/file-integrity-operator/pull/821/changes/89b3f9a97e74d644ad6e5efe20142a7dcd17d73c

Edit: totally wrong link.

.tekton/file-integrity-operator-bundle-dev-pull-request.yaml
Vincent056: I think bundle should always be built as long as we have triggered operator build in a PR.

yuumasato: Yes, that is why we need to have this changed in the operator image pipeliene.

The bundle pipeline may not always have an operator image to wait for if only the bundle image was triggered.
A bundle build should not require an operator build to happen as well.

PR #823, opened 2025-12-11
Separate build trigger for operator and bundle

yuumasato:
Let's not build both images all the time.
This leads to two snapshots whenever there is a build, and makes it a bit more difficult to find the correct one for release.


Vincent056 (commented):
could we use a different tag for the bundle image here

yuumasato:
> could we use a different tag for the bundle image here

What is the motivation?

yuumasato:
Triggering the operator pipeline for bundle only related changes also lead to infinite nudges.
See https://github.com/openshift/file-integrity-operator/pull/851, which was opened after https://github.com/openshift/file-integrity-operator/pull/850 merged.

Vincent056 (approved):
/lgtm

PR #847, opened 2025-12-15
CMP-3954: Restrict gomod updates on release branches

yuumasato:
We are more cautious about updating modules in release branches.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,22 @@
 {
+  "extends": [
+    "config:recommended"
+  ],
   "labels": ["ok-to-test","px-approved","docs-approved","qe-approved"],
   "schedule": [
       "every weekend"
-  ]
+  ],
+  "gomod": {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,27 @@
 {
-  "labels": ["ok-to-test","px-approved","docs-approved","qe-approved"],
+  "extends": [
+    "config:recommended"
+  ],
+  "labels": [
+    "ok-to-test",
+    "px-approved",
+    "docs-approved",
+    "qe-approved"
```
rhmdnd, line 9: We will still need to come through an /lgtm the PR so it doesn't land automatically, right?

yuumasato: @rhmdnd Yes, see for example https://github.com/openshift/file-integrity-operator/pull/828

rhmdnd (approved):
/lgtm

PR #789, opened 2025-11-20
CMP-3954: Add redhat.repo file

rhmdnd:
This will make it so Konflux keeps the RPM lock file up to date, reducing the likelihood that we'll get stale RPMs when we do builds (reducing manualy work).

Upstream commit https://github.com/openshift/file-integrity-operator/pull/787

PR #926, opened 2026-05-04
Enable RPM refreshes from MintMaker/Renovate

yuumasato:
Add source repo file and rpms.in.yaml file so that Renovate can auto refresh RPMs for us.

Based on #787, #789, #790.
The `rpms.lock.yaml` was not create with `--bare` option, so the list of RPMs is shorter.


rhmdnd (approved):
/lgtm

PR #953, opened 2026-05-11
CMP-4285: Use lowercase "https" scheme for ServiceMonitor compatibility

rhmdnd:
The prometheus-operator constant monitoring.SchemeHTTPS resolves to
uppercase "HTTPS", which is accepted by prometheus-operator >= 0.87.0
but rejected by older CRD versions that only validate lowercase "http"
and "https". This causes the metrics ServiceMonitor to fail creation on
clusters running older prometheus-operator CRDs.

Use monitoring.Scheme("https") instead for compatibility across all
supported OpenShift versions. The endpoint configuration logic is
extracted into configureMetricsEndpoints() with a unit test that will
catch any regression back to SchemeHTTPS.

Relevant upstream change: https://github.com/prometheus-operator/prometheus-operator/pull/8017

### DIFF
--- a/cmd/manager/operator.go
+++ b/cmd/manager/operator.go
@@ -394,12 +394,42 @@ func createServiceMonitor(ctx context.Context, cfg *rest.Config, mClient *moncli
 		return nil
 	}
 
-	serverName := fmt.Sprintf("metrics.%s.svc", namespace)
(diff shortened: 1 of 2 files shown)


Vincent056 (approved):
/lgtm

PR #999, opened 2026-07-06
RFE-9579: Add support for resources in FileIntegrity

eloy-valderacarrasco:
## Add configurable resources for the AIDE daemon (`spec.resources`)

Adds an optional `spec.resources` field to the `FileIntegrity` CR so you can set CPU/memory requests and limits for the AIDE daemon pods.

### Why?

The AIDE daemon's memory and CPU usage scales with the size of the filesystem it scans. On large clusters — nodes with many files, a large `/`, or dense workloads — the fixed default limits are too low: the daemon gets OOMKilled or CPU-throttled, so integrity scans slow down, fail, or never complete, leaving nodes without reliable file-integrity monitoring.

Right now there's no way to raise these limits. Making resources configurable lets cluster admins right-size the daemon per environment — increasing limits on large/busy clusters, or tightening them on small ones — instead of being stuck with one-size-fits-all values.

### What changed

**API:**
- Added optional `resources` field to the top-level `FileIntegrity` spec (`spec.resources`)
- Set built-in defaults via a `+kubebuilder:default` marker (requests: 40Mi/40m, limits: 600Mi/300m)
- Updated CRDs and bundle manifests

**Controller:**
- Applies `spec.resources` to the AIDE daemon container at DaemonSet create time
- Detects changes and restarts the DaemonSet when resources are updated
- Falls back to built-in defaults when the field is unset

**Tests:**
- E2E test for default resources and for updating them (`TestFileIntegrityChangeResources`)

**Docs:**
- Updated README with example usage and the new `resources` field
- Updated CRD field descriptions

### Example

```yaml
apiVersion: fileintegrity.openshift.io/v1alpha1
kind: FileIntegrity
metadata:
  name: example-fileintegrity
  namespace: openshift-file-integrity
spec:
  resources:
    requests:
      memory: "100Mi"
      cpu: "100m"
    limits:
      memory: "2Gi"
      cpu: "1"
  config: {}
```


pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -666,6 +667,20 @@ func updateDSPriorityClassName(currentDS *appsv1.DaemonSet, fi *v1alpha1.FileInt
...
 
+// Returns true when the daemon container resources derived from the FileIntegrity
+// object differ from the current DS. Returns false if there was no difference.
+// If an update is needed, this will update the resources from the given DaemonSet.
+func updateDSResources(currentDS *appsv1.DaemonSet, fi *v1alpha1.FileIntegrity, logger logr.Logger) bool {
+	resRef := &currentDS.Spec.Template.Spec.Containers[0].Resources
+	expectedResources := getDaemonResources(fi)
+	needsUpdate := !reflect.DeepEqual(*resRef, expectedResources)
+	if needsUpdate {
+		logger.Info("FileIntegrity needed daemon resources update")
```
parametalol:
```suggestion
		logger.Info("FileIntegrity needed DaemonSet resources update")
```

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -1064,6 +1070,27 @@ func getDebug(fi *v1alpha1.FileIntegrity) string {
...
+	return corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("40Mi"),
+			corev1.ResourceCPU:    resource.MustParse("40m"),
+		},
+		Limits: corev1.ResourceList{
+			corev1.ResourceMemory: resource.MustParse("600Mi"),
+			corev1.ResourceCPU:    resource.MustParse("300m"),
+		},
+	}
```
parametalol, line 1091: nit: extract the default value to a global variable?

eloy-valderacarrasco: Thanks for your comments, I've made the changes requested

PR #1003, opened 2026-07-20
updated go and vendors

vickeybrown:
- Go updates to fix CVEs and updates that the Konflux bot had opened


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/file-integrity-operator
```
Vincent056, line 2: I wonder if we want to keep it as

```
go 1.25.11
```
for this release

yuumasato (approved):
/lgtm

PR #1004, opened 2026-07-22
Add golangci-lint CI workflow and Makefile target

parametalol:
## Summary

- Add a GitHub Actions workflow (`golangci-lint.yaml`) that runs golangci-lint on pushes and PRs to master, using `golangci/golangci-lint-action@v9` with golangci-lint v2.12
- Add a `make lint` target for running golangci-lint locally
- Add a minimal `.golangci.yml` config with vendor mode and the standard linter set (errcheck, govet, staticcheck, unused, ineffassign, gosimple)

## Test plan

- [ ] Verify the GitHub Actions workflow triggers on a PR to master
- [x] Run `make lint` locally and confirm it invokes golangci-lint correctly
- [x] Confirm vendor directory is not linted

PR #1005, opened 2026-07-27
Fix e2e-aws flake cascade and metrics-pod connection race

parametalol:
## Summary

PRs #1003 and #1004 both failed `e2e-aws` despite neither touching e2e tests, RBAC, or the metrics endpoint. Both failed for the same underlying reason:

- `tests/framework/resource.go`: `InitializeClusterResources` hard-fails if a cluster-scoped RBAC object (the `file-integrity-operator` ClusterRole/ClusterRoleBinding) already exists. Since `E2E_SKIP_CLEANUP_ON_ERROR` defaults to `true`, one failing test leaves these behind, and every subsequent test in the serial run then fails immediately at setup instead of just the one that actually broke. `createFromYAML` already supports `skipIfExists` for exactly this; the only caller just never used it. This PR flips it to `true`.
- `tests/e2e/helpers.go`: `runOCandGetOutput` (used to curl the metrics endpoint from an ephemeral pod) fails the test on the very first error. In practice curl exit 7 ("connection refused") happens routinely while the metrics service is still coming up — it shows up 11 times in a fully passing run I checked, just non-fatally, because that other call site already retries via a `wait.Poll` loop. This PR adds the same tolerance (retry a few times, 5s apart) directly in `runOCandGetOutput` so the ~10 tests that go through it aren't one-shot.

Together these mean a single transient hiccup stays isolated to (at most) one test instead of taking down the rest of the suite.

## Test plan

- [x] `go build ./tests/...`
- [x] `go vet ./tests/...`
- [ ] `e2e-aws` Prow job passes (can't run locally — requires a live OpenShift cluster)

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -2449,34 +2449,30 @@ func runOCandCheckError(t *testing.T, args []string) error {
 	return nil
 }
 
-// runOCGetOutputMaxAttempts and runOCGetOutputRetryDelay bound the retries in runOCandGetOutput.
-// The ephemeral pods it drives routinely hit curl exit 7 (connection refused) when the target
-// service isn't quite ready yet, which is expected and recovers within a few seconds.
-const (
-	runOCGetOutputMaxAttempts = 3
-	runOCGetOutputRetryDelay  = 5 * time.Second
-)
-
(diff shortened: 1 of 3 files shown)


abushkin-redhat:
Hi Michaël, 

The PR looks good. /lgtm

PR #845, opened 2025-12-15
CMP-3757: Add retry logic for operator metrics secrets

yuumasato:
Fixes crash-restart cycle when metrics secrets are not yet synced by service-ca and token controllers. 
Uses exponential backoff with 10 retries, matching the daemon's approach for BuildConfigFromFlags.

PR #1010, opened 2026-07-29
Fix controller metrics server race with serving-cert volume sync

Vincent056:
## What

Makes the custom `:8585` controller metrics server wait (with exponential backoff, up to 5 minutes) for the serving-cert key pair to be synced into the container's volume mount before calling `ListenAndServeTLS`, propagates server failures to the manager instead of swallowing them, and shuts the server down cleanly when the manager context is canceled.

## Why

`TestMetricsHTTPVersion` fails deterministically on e2e-aws (e.g. on #1003, and its skipped cleanup cascades into `clusterrole already exists` failures for every subsequent test). The failing curl gets exit 7 (connection refused) against `https://metrics.<ns>.svc:8585/metrics-fio`, and the operator log shows why:

```
{"logger":"metrics","msg":"Metrics service failed",
 "error":"open /var/run/secrets/serving-cert/tls.crt: no such file or directory"}
```

This is a regression from #845: before it, the operator exited and restarted when the serving-cert secret was missing, and by the time the container came back the kubelet had synced the cert files into the mount. Now the operator waits in-process — but only for the secret to exist **in the API**. `Metrics.Start` then does a one-shot `ListenAndServeTLS` before the kubelet has synced the files into the volume, fails, logs, and returns `nil`, so the metrics server never comes up and nothing restarts.

The fix mirrors the backoff pattern #845 introduced for the API-side checks, applied to the mounted files. Returning the error after the backoff is exhausted restores the old exit-and-restart behavior as a last resort instead of silently running without metrics.

### DIFF
--- a/pkg/controller/metrics/metrics.go
+++ b/pkg/controller/metrics/metrics.go
@@ -5,7 +5,10 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"os"
+	"time"
 
+	backoff "github.com/cenkalti/backoff/v4"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


yuumasato (approved):
/lgtm

yuumasato:
Evidence that the backoff worked: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_file-integrity-operator/1010/pull-ci-openshift-file-integrity-operator-master-e2e-aws/2082552773961846784/artifacts/e2e-aws/test/artifacts/e2e-test-logcompress_file-integrity-operator-645f874644-76r7r_self.log

yuumasato:
Manually merging so we can move forward with FIO v1.4.0 release.

PR #1006, opened 2026-07-27
Fix e2e-aws metrics-test PodSecurity violations and flake cascade

vickeybrown:
## Summary

Supersedes #1005 — includes all of @parametalol's changes plus one additional fix:

- **Retry logic for `runOCandGetOutput`** — ephemeral metrics pods routinely hit curl exit 7 (connection refused) while the metrics service is coming up; retries with `wait.PollImmediate` instead of failing on the first attempt (from #1005)
- **Replace stale cluster-scoped resources** — `InitializeClusterResources` now deletes and recreates cluster-scoped RBAC objects left behind by a previous test whose cleanup was skipped, instead of hard-failing on "already exists" (from #1005)
- **Add PodSecurity `restricted:latest` security context to `metrics-test` pods** — all four `oc run` call sites now include `--overrides` setting `runAsNonRoot`, `seccompProfile`, `allowPrivilegeEscalation=false`, and `capabilities.drop=["ALL"]` so the ephemeral pods comply with enforced PodSecurity policy

## Test plan

- [x] `go build ./tests/...`
- [x] `go vet ./tests/...`
- [ ] `e2e-aws` Prow job passes

Co-Authored-By: Michaël <github@parameta.lol>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1011, opened 2026-07-31
Fix e2e-aws metrics-test pod crash and PruneBackup flake

yuumasato:
## Summary

- **Fix metrics-test pod crash**: The `oc run --overrides` added in #1006 set `runAsNonRoot: true` at the pod level but omitted `runAsUser`. Since `fedora-minimal` runs as root (UID 0) by default, the kubelet refused to start the container, causing every test that scrapes the metrics endpoint to get empty output and fail. Adding `runAsUser: 65534` (nobody) satisfies the constraint.

- **Fix TestFileIntegrityPruneBackup flake**: The test replaced the entire `FileIntegrityConfig` struct to set `MaxBackups=1`, which zeroed `GracePeriod`. The CRD then defaulted it to 900s, changing the daemon args and triggering an unnecessary DaemonSet rolling restart. The reinit could race with the rollout, leaving old daemon pods (`MaxBackups=5`) to handle the reinit without pruning. Setting only `MaxBackups` on the existing config avoids the spurious rollout.

## Test plan

- [ ] e2e-aws passes — all 7 metrics-related test failures should be resolved
- [ ] TestFileIntegrityPruneBackup no longer flakes from the GracePeriod/rollout race

### DIFF
--- a/tests/e2e/helpers.go
+++ b/tests/e2e/helpers.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


tests/e2e/helpers.go
```diff
@@ -2395,9 +2413,10 @@ func logContainerOutput(t *testing.T, f *framework.Framework, namespace, name st
 }
 
 func getMetricResults(t *testing.T, namespace string) string {
+	const image = "registry.fedoraproject.org/fedora-minimal:latest"
```
parametalol, line 2416: What about pinning the image to a sha?

E.g.: latest (today) = 44 (today) = `sha256:fd7b2078b72a40582cbd14aa341551d953fe8a2f79350f6c426438b1a8fbd5c4`

yuumasato:
`e2e-aws` passed, re-running to check if get a consistent pass

Vincent056 (approved):
/lgtm

tests/e2e/helpers.go
yuumasato: @parametalol What are your concerns with using `latest` tag?

parametalol: Rolling latest tag may introduce an unexpected and unrelated instability, harder to identify and reproduce if the tag silently shifts to a different version.

PR #1038, opened 2026-08-04
make bundle: update all files with specified version

yuumasato:
Now, running `make bundle VERSION=x.y.z` bumps all the required files.
Running `make bundle` just regenerates the bundles while keeping the current version.


Vincent056 (approved):
/lgtm

PR #1041, opened 2026-08-05
Pin AIDE version to 0.16

yuumasato:
The config we are using in FIO is compatible with AIDE 0.16.

Migrating to 0.18 or 0.19 will require migration steps.


yuumasato:
Follow up from https://github.com/openshift/file-integrity-operator/pull/1042

Vincent056 (approved):
/lgtm

PR #1045, opened 2026-08-06
Re-Add replaces field to CSV and update Makefile to preserve it

yuumasato:
The update-skip-range Makefile target was removing the replaces field from the base CSV.

Replace the deletion with a conditional update that automatically sets replaces to the previous version when VERSION changes, and leaves it untouched otherwise.
Also update the base CSV version during bundle generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -297,7 +297,11 @@ endif
 
 .PHONY: update-skip-range
 update-skip-range: check-operator-version
-	sed -i '/replaces:/d' config/manifests/bases/file-integrity-operator.clusterserviceversion.yaml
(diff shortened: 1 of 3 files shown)


yuumasato:
The Makefile changes have been incorporated into https://github.com/openshift/file-integrity-operator/pull/1038 for `master`.

yuumasato:
Sorry, as pointed out by Vincent, downstream's manifest is updated by https://github.com/openshift/file-integrity-operator/blob/8b647112852d8dc3dbbcd5189cac2a82c456a923/bundle-hack/update_csv.go#L98 so the replaces key should be there in the current 1.4.1 build

PR #1059, opened 2026-09-02
CMP-4638: central TLS profile consistency

parametalol:
Heavily influenced by ComplianceAsCode/compliance-operator#1373.

Dependencies bump is also in a separate #1058.


Vincent056:
The bundle CSV was not regenerated after adding `config.openshift.io/apiservers get,list,watch` to `config/rbac/operator_clusterrole.yaml`. `bundle/manifests/file-integrity-operator.clusterserviceversion.yaml` has no `config.openshift.io` rule, and `bundle.Dockerfile` copies `bundle/manifests` verbatim, so `ci-index-file-integrity-operator-bundle` / `e2e-bundle-aws*` would install the operator without the permission.

I reproduced what that does on a 4.21 cluster by removing the `apiservers` rule from the ClusterRole and restarting the pod:

```
"Could not fetch APIServer TLS profile, using defaults" ... apiservers.config.openshift.io "cluster" is forbidden
"Failed to watch" reflector ... failed to list *v1.APIServer: apiservers.config.openshift.io is forbidden   (repeats)
"Could not wait for Cache to sync" controller="tlssecurityprofilewatcher" ... timed out waiting for cache to be synced
"Manager exited non-zero" error="failed to wait for tlssecurityprofilewatcher caches to sync ..."
```

The container exits 1 after the 2 minute cache-sync timeout and crash-loops. The "using defaults" fallback in `fetchClusterTLSSettings` does not help because the watcher registered in `makeSecurityProfileWatcher` is fatal on cache-sync failure, so the RBAC is a hard runtime requirement.

Please run `make bundle` and commit `bundle/manifests`, then confirm `e2e-bundle-aws` is green (it has not run yet because of the Go 1.26 build failure). Optionally consider whether the watcher should degrade to "no change detection" instead of taking the operator down when the permission is missing (e.g. during an upgrade where the CSV RBAC lands late); CO made the same fatal choice, so at minimum it should be documented.

yuumasato:
Not sure if the prow tests are running or not, it has been 2 days since last change. Results should have been reported by now.
/retest

yuumasato:
Superseded by https://github.com/openshift/file-integrity-operator/pull/1065

PR #1063, opened 2026-09-07
CMP-4573: switch runtime base image to ubi9-minimal-pqc

yuumasato:
Builds upon #1057 and updates the RPM lock files.

Test whether Konflux can build it now.
Regenerated the RPM lock file with `--bare` option so that `libgcrypt` is included.

### DIFF
--- a/build/Dockerfile.openshift
+++ b/build/Dockerfile.openshift
@@ -12,7 +12,7 @@ COPY . .
 
 RUN make build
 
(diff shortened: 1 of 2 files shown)


guzalv (approved):

parametalol (approved):

Vincent056 (approved):
/lgtm

PR #1064, opened 2026-09-17
CMP-4728: Add support for pod labels and annotations in FileIntegrity

taimurhafeez:
## What

Adds optional `labels` and `annotations` fields to the `FileIntegrity` spec. When set, they are applied to the pod template of the AIDE daemonSet:

```yaml
apiVersion: fileintegrity.openshift.io/v1alpha1
kind: FileIntegrity
metadata:
  name: example-fileintegrity
spec:
  labels:
    example.com/team: security
  annotations:
    example.com/scrape: "true"
  nodeSelector:
    node-role.kubernetes.io/worker: ""
```

## Why

Observability stacks commonly onboard workloads via pod label-based autodiscovery. The daemon pods carry only operator-generated labels and the CRD exposed no way to add custom ones, leaving a security-critical component invisible to monitoring pipelines. The only workaround was hand-editing the operator-managed daemonSet, which reconciliation immediately reverts.

## Implementation notes

- The fields apply both on daemonSet creation and when an existing `FileIntegrity` is edited. The latter goes through the existing `updateDS*` drift-correction path, which is easy to miss — a field wired only into `aideDaemonset()` would silently never apply to an existing CR.
- The whole map is replaced rather than merged, so removing a key from the CR also removes it from the pod template. This matches how `nodeSelector` and `tolerations` already behave. Worth knowing: annotations added to the pod template by other controllers will be stripped on the next reconcile.
- Custom labels go on the pod template only, never on `spec.selector`, which is immutable — so this applies in place to already-deployed daemonSets rather than forcing a delete/recreate. There is a unit test asserting the selector stays untouched.
- **Metadata-only changes trigger a DaemonSet rolling update** (maxUnavailable: 1, node-by-node) **rather than killing all AIDE pods at once**. The existing force-restart path (`RestartFileIntegrityDs`) is reserved for changes that require it (args, config, volumes, scripts).
- **Operator-managed keys are rejected at admission** via CEL: `app` is the daemonSet selector and the selector `RestartFileIntegrityDs` uses, and the `file-integrity.openshift.io/` prefix covers the marker the operand NetworkPolicies select on plus ownership tracking. `podTemplateLabels` also lets operator keys win as a defence-in-depth safety net.
- **Format validation** uses CEL for common errors (first character, size limits) within the API server's cost budget. Full Kubernetes label/annotation format validation is performed by the DaemonSet API. Invalid formats that pass admission are caught at DaemonSet creation/update and **surfaced as Warning Events** on the FileIntegrity object, so `oc describe` shows why labels/annotations did not apply.
- **Runtime-affecting annotation prefixes are blocked** at admission: `k8s.v1.cni.cncf.io/`, `io.kubernetes.cri-o.`, `container.apparmor.security.beta.kubernetes.io/`, `seccomp.security.alpha.kubernetes.io/`. These can break the privileged DaemonSet (e.g., a non-existent `NetworkAttachmentDefinition` leaves all pods `Pending`). The field is for observability metadata.
- Both maps are bounded to 64 entries. The API server estimates CEL rule cost from the schema's maximum size, and an unbounded map can push an `all()` rule over the per-rule budget, which would get the whole CRD rejected at apply time.
- `podTemplateAnnotations` returns `nil` rather than an empty map when unset. An empty non-nil map is not `reflect.DeepEqual` to `nil`, which would make the drift helper report a change on every reconcile and restart the daemon pods each pass.
- The transient re-init daemonSet is deliberately left untouched.
- **DaemonSet create/update failures are surfaced as Warning Events** on the FileIntegrity object, so users see validation errors in `oc describe fileintegrity`.

## Testing

- **Unit** (`make test-unit`): 15 subtests covering the metadata builders, the merge/precedence rules, the nil-on-empty behaviour, the selector invariant, and the drift helpers for add/change/remove/no-op.
- **E2E**: `TestFileIntegrityPodLabelsAndAnnotations` covers creation, then an update that simultaneously changes, adds and removes keys, verification that **running pods carry the metadata** (not just the pod template), then re-asserts the scan returns to `Active`.
- **Manual, on OCP 4.22.14:** verified the metadata reaches the running pods and is selectable by label; that an update converges in place (`generation: 2`, unchanged selector) via **rolling update over 2m47s** (6 pods restarted 33-34s apart, not all-at-once); that reserved keys and invalid first characters are rejected at admission; that runtime-affecting annotations (`k8s.v1.cni.cncf.io/networks`) are blocked; that invalid formats passing admission (e.g., `"not valid!"`) fail at DaemonSet creation with a **Warning Event** showing the full validation error; that scans stay `Active` with all nodes `Succeeded` and the operand NetworkPolicies still select every pod; and that removing both maps leaves the daemonSet generation stable with no repeated drift, confirming there is no reconcile hot-loop.

## Review feedback addressed

**Vincent056's findings:**

1. **Invalid formats crashed operator silently** → Fixed with Warning Events on FileIntegrity object (`oc describe` shows DaemonSet validation failures). CEL format validation added for common errors (first char, size limits); full regex validation exceeded cost budget by 2-24x even with `maxProperties=64`. The two-layer approach (admission + DaemonSet API + Warning Events) ensures users always see errors.

2. **Metadata-only changes deleted all pods at once** → Fixed. Labels/annotations changes now trigger DaemonSet rolling update only. Config/args/volume changes still force-restart for immediate pickup. Validated: 6 pods restarted 33-34s apart over 2m47s (vs all in same second before).

3. **Dangerous annotations broke DaemonSet** → Fixed. All four runtime-affecting prefixes blocked at admission with clear error messages.

4. **Documentation clarity** → Updated: "AIDE daemon pods" (not "operator pods"), re-init exclusion noted, replace semantics documented, restart behavior explained.

5. **E2E test gap** → Extended test to verify running pods carry labels/annotations, not just the DaemonSet template.

Ref: https://issues.redhat.com/browse/CMP-4728  
Ref: https://issues.redhat.com/browse/CMP-4551


Vincent056:
Validated this on a 4.21 cluster with the PR build (create, update, removal, CEL, empty maps, upgrade from a master-built DaemonSet, re-init, NetworkPolicies), twice. The feature works as described and the generated files, unit tests and selector invariant are all in order. Three things I would like changed before merge, two of them small:

1. **Invalid keys/values pass the CRD and then fail every reconcile silently.** `labels: {example.com/team: "not valid!"}` or a key `bad key!` is accepted, after which the DaemonSet update is rejected (`spec.template.labels: Invalid value: "not valid!": a valid label must be ...`), the reconciler errors 12-13 times per 20 s with backoff, the pods keep the old template, and the FileIntegrity stays `Active` with no condition or Event. Please validate the format at admission (on this 4.21 cluster `format.qualifiedName().validate(k).hasValue()` / `format.labelValue().validate(self[k]).hasValue()` work; if 1.5.0 must support OCP older than 4.18, a `matches()` regex does the same), and surface DaemonSet update failures on the object (condition or Warning Event).

2. **Metadata-only changes kill all AIDE pods at once.** The new flags feed the existing `Update` + `RestartFileIntegrityDs` block, so a label edit triggers the DaemonSet rolling update and then a delete-all; all three pods were recreated in the same second on every change, and the same happens when the operator strips a foreign template annotation (31 s after `oc rollout restart`, for example). For labels/annotations the pod template change alone is enough; suggest leaving `RestartFileIntegrityDs` out of the metadata-only case.

3. **Annotations are passed through unfiltered to a privileged DaemonSet.** `k8s.v1.cni.cncf.io/networks: <nad>` on the FileIntegrity attaches the AIDE pods to another network; with a non-existent NAD all pods sat in `Pending` (`cannot find a network-attachment-definition`) while the FileIntegrity showed `Initializing` and the node statuses stayed `Succeeded`. Either reject the runtime-affecting prefixes (`k8s.v1.cni.cncf.io/`, `io.kubernetes.cri-o.`, `container.apparmor.security.beta.kubernetes.io/`, ...) or document clearly that the field is for observability metadata and the operator does not validate runtime effects; item 1's status feedback covers the rest.

Smaller: the CRD/README say "pods created by the operator" but the re-init DaemonSet is deliberately excluded, please say "AIDE daemon pods"; document the replace (not merge) semantics and the restart; title should reference the story [CMP-4728](https://redhat.atlassian.net/browse/CMP-4728) rather than the epic; and please rebase, the Konflux `prefetch-dependencies` failure comes from the base commit `92cf1117` (its own on-push pipelines failed the same way) and master is fixed since #1066. The prow e2e jobs have not run yet for this PR; I ran `TestFileIntegrityPodLabelsAndAnnotations` against my cluster with the PR image (result in the follow-up).

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -53,6 +53,18 @@ type FileIntegritySpec struct {
 	// This is an optional field. If PriorityClass is invalid or not found,
 	// it will be ignored and cleared from the spec.
 	PriorityClassName string `json:"priorityClassName,omitempty"`
+	// Labels to add to the pods created by the operator. Keys must not use the
+	// reserved key "app" or the reserved "file-integrity.openshift.io/" prefix,
+	// which are managed by the operator.
+	// +kubebuilder:validation:MaxProperties=64
+	// +kubebuilder:validation:XValidation:rule="self.all(k, k != 'app' && !k.startsWith('file-integrity.openshift.io/'))",message="labels must not use the reserved key 'app' or the reserved prefix 'file-integrity.openshift.io/'"
```
Vincent056, line 63: This rule only rejects reserved keys; key/value *format* is not validated. On a 4.21 cluster `labels: {example.com/team: "not valid!"}` and `labels: {"bad key!": v}` are both accepted, and the operator then fails every reconcile with `DaemonSet.apps "aide-<name>" is invalid: spec.template.labels: Invalid value: "not valid!": a valid label must be ...` (12-13 reconciler errors per 20 s with backoff), while the FileIntegrity stays `Active` with no condition or Event.

Please validate the format at admission. Verified on this cluster (k8s 1.34): `format.qualifiedName().validate(k).hasValue()` and `format.labelValue().validate(self[k]).hasValue()` reject both cases and accept `valid-1`. The CEL `format` library is available from Kubernetes 1.32 (OCP 4.19+); if 1.5.0 must still install on older OCP, a `k.matches(...)`/`self[k].matches(...)` regex works everywhere. `maxLength: 63` on values would be good too.

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -53,6 +53,18 @@ type FileIntegritySpec struct {
...
+	Labels map[string]string `json:"labels,omitempty"`
+	// Annotations to add to the pods created by the operator. Keys must not use
+	// the reserved "file-integrity.openshift.io/" prefix, which is managed by
+	// the operator.
+	// +kubebuilder:validation:MaxProperties=64
+	// +kubebuilder:validation:XValidation:rule="self.all(k, !k.startsWith('file-integrity.openshift.io/'))",message="annotations must not use the reserved prefix 'file-integrity.openshift.io/'"
```
Vincent056, line 77: Same format point for annotation keys (`bad key!` is accepted here too). Separately: these annotations land on a privileged DaemonSet with the host root mounted, and some are not inert. `k8s.v1.cni.cncf.io/networks: <nad>` attaches the AIDE pods to another network; with a non-existent NAD all three pods sat in `Pending` (`cannot find a network-attachment-definition`) while the FileIntegrity showed `Initializing` and the node statuses stayed a stale `Succeeded`. Consider rejecting the runtime-affecting prefixes (`k8s.v1.cni.cncf.io/`, `io.kubernetes.cri-o.`, `container.apparmor.security.beta.kubernetes.io/`, `seccomp.security.alpha.kubernetes.io/`), or at least document that the field is for observability metadata and is not validated for runtime effects.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -600,8 +600,11 @@ func (r *FileIntegrityReconciler) FileIntegrityControllerReconcile(request recon
 		tolsNeedsUpdate := updateDSTolerations(dsCopy, instance, reqLogger)
 		pcNeedsUpdate := updateDSPriorityClassName(dsCopy, instance, reqLogger)
 		volsNeedUpdate := updateDSContainerVolumes(dsCopy, instance, operatorImage, reqLogger)
+		labelsNeedUpdate := updateDSLabels(dsCopy, instance, reqLogger)
+		annotationsNeedUpdate := updateDSAnnotations(dsCopy, instance, reqLogger)
 
-		if argsNeedUpdate || imgNeedsUpdate || nsNeedsUpdate || tolsNeedsUpdate || pcNeedsUpdate || volsNeedUpdate || scriptsUpdated {
+		if argsNeedUpdate || imgNeedsUpdate || nsNeedsUpdate || tolsNeedsUpdate || pcNeedsUpdate ||
+			volsNeedUpdate || labelsNeedUpdate || annotationsNeedUpdate || scriptsUpdated {
 			if err := r.Client.Update(context.TODO(), dsCopy); err != nil {
```
Vincent056, line 610: When this `Update` is rejected (invalid label key/value from the CR, see the types.go comment), the error is only returned and logged; nothing reaches the object. Suggest recording a Warning Event on the FileIntegrity (the reconciler already has `r.Recorder`) or a status condition, so `oc describe fileintegrity` explains why the new labels/annotations did not apply.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -600,8 +600,11 @@ func (r *FileIntegrityReconciler) FileIntegrityControllerReconcile(request recon
 		tolsNeedsUpdate := updateDSTolerations(dsCopy, instance, reqLogger)
 		pcNeedsUpdate := updateDSPriorityClassName(dsCopy, instance, reqLogger)
 		volsNeedUpdate := updateDSContainerVolumes(dsCopy, instance, operatorImage, reqLogger)
+		labelsNeedUpdate := updateDSLabels(dsCopy, instance, reqLogger)
+		annotationsNeedUpdate := updateDSAnnotations(dsCopy, instance, reqLogger)
 
-		if argsNeedUpdate || imgNeedsUpdate || nsNeedsUpdate || tolsNeedsUpdate || pcNeedsUpdate || volsNeedUpdate || scriptsUpdated {
+		if argsNeedUpdate || imgNeedsUpdate || nsNeedsUpdate || tolsNeedsUpdate || pcNeedsUpdate ||
+			volsNeedUpdate || labelsNeedUpdate || annotationsNeedUpdate || scriptsUpdated {
```
Vincent056, line 609: Because `labelsNeedUpdate`/`annotationsNeedUpdate` feed this block, a metadata-only edit goes through `RestartFileIntegrityDs` below and deletes every AIDE pod at once, on top of the DaemonSet's own RollingUpdate for the template change. Observed on every label edit (the operator logs `daemon configuration changed - pods restarted`); in three of four runs all three pods were recreated in the same second. The delete-all exists for args/config/volume changes; for pod-template labels and annotations the DaemonSet rolling update (maxUnavailable 1) is enough and node-by-node. Suggest keeping the two metadata flags in the `Update` condition but leaving them out of the restart condition. The same double restart happens when the operator reverts a foreign annotation on the pod template (measured about 30 s after patching one in; `oc rollout restart ds/aide-<name>` adds one the same way).

Vincent056 (commented):
Inline follow-ups to my comment above (cluster-validated on 4.21, each point reproduced twice). No suggestion blocks on purpose: the CEL fix depends on the minimum OCP the 1.5.0 bundle targets.

README.md
```diff
@@ -64,6 +68,8 @@ In the `spec`:
 * **nodeSelector**: Selector for nodes to schedule the scan instances on.
 * **tolerations**: Specify tolerations to schedule on nodes with custom taints. When not specified, a default toleration allowing running on master and infra nodes is applied.
 * **priorityClassName**: (Optional) Specifies the `PriorityClass` for the pods created by the operator. If the PriorityClass is invalid or not found, it will be ignored and cleared from the spec.
+* **labels**: (Optional) Custom labels to add to the pods created by the operator, useful for label-based autodiscovery by monitoring and log collection systems. The key `app` and the `file-integrity.openshift.io/` prefix are reserved for the operator and are rejected.
```
Vincent056: Docs nit, applies to the CRD descriptions too: the re-init DaemonSet (`aide-ini-<name>`) deliberately does not get these (verified), so "pods created by the operator" should be "AIDE daemon pods", with a note that re-init pods are excluded. It would also help to state that the maps replace (not merge) whatever is on the pod template, and that changing them restarts the AIDE pods.

tests/e2e/helpers.go
```diff
@@ -1086,6 +1086,141 @@ func setupPriorityClassTest(t *testing.T, integrityName string) (*framework.Fram
 	return f, testctx, namespace
 }
 
+// verifyDaemonSetPodMetadata polls the DaemonSet until its pod template carries every label and
+// annotation in expectedLabels/expectedAnnotations, carries none of the keys in absentLabels, and
+// still carries the labels the operator manages. Polling is needed because the operator
+// reconciles the DaemonSet asynchronously after the FileIntegrity is updated.
+func verifyDaemonSetPodMetadata(t *testing.T, f *framework.Framework, namespace, dsName, integrityName string,
```
Vincent056, line 1093: Minor: this checks the DaemonSet pod template only. Since the point of the feature is label-based discovery of the running pods, a follow-up check that the pods returned by `-l example.com/team=...` carry the labels/annotations (what the PR body verified manually) would close the loop. FWIW the test passed for me against a 4.21 cluster with the PR image (`--- PASS: TestFileIntegrityPodLabelsAndAnnotations (45.40s)`).

pkg/apis/fileintegrity/v1alpha1/fileintegrity_types.go
```diff
@@ -53,6 +53,32 @@ type FileIntegritySpec struct {
...
+	// operator. Format errors are validated at admission (invalid first characters, size limits)
+	// and by the DaemonSet API (full format validation), with failures surfaced as Warning Events.
+	// Re-init pods are excluded. The entire map replaces (does not merge with) the pod template
+	// labels, and changing labels triggers a rolling restart of the AIDE daemon pods.
+	// +kubebuilder:validation:MaxProperties=64
+	// +kubebuilder:validation:XValidation:rule="self.all(k, k != 'app' && !k.startsWith('file-integrity.openshift.io/'))",message="labels must not use the reserved key 'app' or the reserved prefix 'file-integrity.openshift.io/'"
+	// +kubebuilder:validation:XValidation:rule="self.all(k, size(k) >= 1 && size(k) <= 253 && size(self[k]) <= 63)",message="label keys must be 1-253 characters and values must be at most 63 characters"
+	// +kubebuilder:validation:XValidation:rule="self.all(k, !k.startsWith(' ') && !k.startsWith('-') && !self[k].startsWith(' ') && !self[k].startsWith('-') && (self[k] == '' || !self[k].startsWith('!')))",message="label keys and values must start with an alphanumeric character, not space, '-', or '!'"
```
yuumasato: The reserved-key and runtime-prefix rules are great, keep those. It's the size/first-char format rules I'd reconsider.

We can't do full format validation at admission on this support matrix anyway: `format.labelValue()` needs OCP 4.18+ (we still ship on 4.12/4.14/4.16), and a full `matches()` regex blows the 10M per-rule CEL cost budget since the estimator can't use `size(self[k])` to bound the value length. So these rules only catch *some* format errors early — `!bad` fails at apply, but `b a d` slips through to the DaemonSet API.

Since the DS API + Warning Events already catch everything, I'd just drop the size/first-char rules and let that be the single, consistent validation path. If you'd rather keep the early catches for common typos, fine — just add a one-line comment saying they're best-effort and the DS API is authoritative.

pkg/controller/fileintegrity/fileintegrity_controller.go
```diff
@@ -600,22 +602,36 @@ func (r *FileIntegrityReconciler) FileIntegrityControllerReconcile(request recon
...
 
-			// TODO: We might want to change this to something that signals to the daemonSet pods that they need to
-			// gracefully exit, and let them restart that way.
-			err := common.RestartFileIntegrityDs(r.Client, common.DaemonSetName(instance.Name))
-			if err != nil {
-				return reconcile.Result{}, err
+			// For metadata-only changes (labels/annotations), the DaemonSet's RollingUpdate is
+			// sufficient (maxUnavailable: 1, node-by-node). Only force-restart all pods when the
+			// change requires it (args, config, volumes, scripts).
+			needsRestart := argsNeedUpdate || imgNeedsUpdate || volsNeedUpdate || scriptsUpdated
```
yuumasato: This also stops force-restarting for `nodeSelector`/`tolerations`/`priorityClassName`, not just labels/annotations — they now rely on the DS rolling update instead. Probably fine (all pod-template changes), but it's outside this PR's scope. Can we keep `ns/tols/pc` in `needsRestart` so only labels/annotations are excluded? Or if it's intentional, note it in the PR description.

yuumasato (commented):
Two follow-ups from re-reviewing the latest push (thanks for addressing the earlier round). Both are non-blocking discussion points rather than hard change requests.

Vincent056 (approved):
/lgtm

config/manager/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ resources:
 
 images:
 - name: file-integrity-operator
-  newName: quay.io/file-integrity-operator/file-integrity-operator
-  newTag: latest
+  newName: quay.io/rh-ee-thafeez/file-integrity-operator
+  newTag: cmp4728-review-fixes-2
```
yuumasato: @taimurhafeez Please remove these changes!

yuumasato (changes requested):

yuumasato (approved):
/lgtm

PR #1065, opened 2026-09-18
CMP-4504,CMP-4638: Honor cluster-wide TLS security profile when configured

guzalv:
## Problem

Per the [CMP-4504](https://redhat.atlassian.net/browse/CMP-4504) epic, OpenShift components must "respect cluster-wide TLS settings when configured to do so" instead of hardcoding their own. FIO's metrics server (`pkg/controller/metrics/metrics.go`, `:8585`) hardcoded `MinVersion: tls.VersionTLS12` plus `libgocrypto.SecureTLSConfig()`'s static cipher list, and the webhook server (`cmd/manager/operator.go`, `:9443`, currently unused but still configured) only disabled HTTP/2 with no TLS version/cipher configuration at all — neither consulted `apiserver.config.openshift.io/cluster`'s `.spec.tlsSecurityProfile`/`.spec.tlsAdherence`.

Note: FIO already offers ML-KEM by default (Go 1.25's `X25519MLKEM768`) — that part of the epic was already satisfied and is untouched here.

## Approach

This mostly mirrors compliance-operator's [commit 03932c22](https://github.com/ComplianceAsCode/compliance-operator/commit/03932c22) ("[CMP-4147](https://redhat.atlassian.net/browse/CMP-4147): Use OpenShift API server TLS configuration"):

- `fetchTLSConfig` reads the `APIServer` singleton once at startup.
- The resolved profile's `tls.Config` mutator is computed once (`NewTLSConfigFromProfile`) and shared between the webhook server's `TLSOpts` and the metrics server's `tls.Config` (`Metrics.SetTLSConfigFn`).
- `newTLSProfileWatcher` polls for changes to `apiserver.config.openshift.io/cluster` on a ticker.
- Poll/fetch failures are logged, counted (`cluster_tls_profile_poll_failed` `error_total`), and recorded as a Warning `Event`.
- Register `configv1` in the operator's scheme (was missing) and add RBAC (`get` on `config.openshift.io/apiservers`, cluster-scoped) needed to read the `APIServer` singleton.

No version bump needed for `github.com/openshift/api` or `github.com/openshift/library-go` — both vendored versions already carry `TLSAdherencePolicy`/`ShouldHonorClusterTLSProfile`. Only `github.com/openshift/controller-runtime-common` is newly vendored.

## A note on RBAC generation

FIO's `make manifests` target only passes `paths=./pkg/apis/fileintegrity/v1alpha1` to `controller-gen`, so — verified empirically — it does **not** actually scan the `+kubebuilder:rbac` markers in `pkg/controller/fileintegrity/setup.go` (a pre-existing situation; those markers already don't round-trip through `make manifests` today either). I added the marker there for consistency with the existing convention, but applied the actual RBAC change by hand to `config/rbac/operator_clusterrole.yaml` and mirrored it into the generated `bundle/manifests/file-integrity-operator.clusterserviceversion.yaml`. I did not run `make bundle` since it requires a release `VERSION` and also bumps unrelated skip-range/version metadata — out of scope for this change.

## Testing

Unit + E2E test, verified in a real cluster:

Built the real Go 1.25.11 binary from this branch, grafted it onto the actual shipped release image (`registry.redhat.io/compliance/openshift-file-integrity-rhel8-operator@sha256:627ef29a...`, from the cluster's `redhat-operators` catalog `stable` channel) via `crane mutate --append`, pushed to the cluster's internal registry, and deployed via `make deploy`.

**Scenario 1 — default cluster config (no `tlsSecurityProfile`/`tlsAdherence` set):**

```
$ oc get apiserver cluster -o jsonpath='{.spec}'
{"audit":{"profile":"Default"}}
```

`tls-scanner --pqc-check` (openshift/tls-scanner) against the metrics endpoint:
```
Port 8585:
    TLS 1.3:  SUPPORTED
    ML-KEM:   SUPPORTED
    ML-KEM KEMs: X25519MLKEM768
    TLS Versions: TLSv1.2, TLSv1.3
PQC COMPLIANCE CHECK: PASSED
```
`testssl.sh` detail: `TLS 1.2 offered (OK)`, `TLS 1.3 offered (OK)`, FS ciphers `TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, ECDHE-RSA-AES256-GCM-SHA384, ECDHE-RSA-CHACHA20-POLY1305, TLS_AES_128_GCM_SHA256, ECDHE-RSA-AES128-GCM-SHA256`, `KEMs offered X25519MLKEM768`.

Behavior is byte-for-byte identical to pre-change (previously verified in a separate PQC-readiness pass), confirming this change is a no-op for the current defaults.

**Scenario 2 — cluster opts into strict adherence with a distinctly different profile:**

`TLSAdherence` is a Tech Preview field in 4.22.13 (`FeatureGate` disabled by default), so I enabled it narrowly via `CustomNoUpgrade` (only that one gate) before it would validate on the `APIServer` CRD:
```
$ oc patch featuregate cluster --type=merge -p '{"spec":{"featureSet":"CustomNoUpgrade","customNoUpgrade":{"enabled":["TLSAdherence"]}}}'
$ oc patch apiserver cluster --type=merge -p '{"spec":{"tlsSecurityProfile":{"type":"Modern","modern":{}},"tlsAdherence":"StrictAllComponents"}}'
$ oc get apiserver cluster -o jsonpath='{.spec}'
{"audit":{"profile":"Default"},"tlsAdherence":"StrictAllComponents","tlsSecurityProfile":{"modern":{},"type":"Modern"}}
```

Operator pod restarted (observed via `oc get events`: `Killing` → new pod `Started`). Note the cluster's own control plane (etcd/kube-apiserver/kube-scheduler/openshift-apiserver) also rolled new static-pod revisions as a result of the cluster-wide TLS profile change — exactly the kind of "many components will restart, plan a maintenance window" behavior the epic calls out; all cluster operators returned to healthy (`Available=True, Degraded=False`) afterward.

After the restart, `tls-scanner`/`testssl.sh` against the same metrics endpoint:
```
 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    not offered
 TLS 1.2    not offered
 TLS 1.3    offered (OK): final
 10.129.2.21:8585 appeared to support TLS 1.3 ONLY.

 FS is offered (OK)           TLS_AES_256_GCM_SHA384
                              TLS_CHACHA20_POLY1305_SHA256
                              TLS_AES_128_GCM_SHA256
 KEMs offered                 X25519MLKEM768
```
```
Port 8585:
    TLS 1.3:  SUPPORTED
    ML-KEM:   SUPPORTED
    ML-KEM KEMs: X25519MLKEM768
    TLS Versions: TLSv1.3
PQC COMPLIANCE CHECK: PASSED
```

TLS 1.2 is now rejected and the cipher list is exactly the 3 ciphers of `configv1.TLSProfiles[TLSProfileModernType]` (`TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384`, `TLS_CHACHA20_POLY1305_SHA256`, min version `VersionTLS13`) — confirming the operator now honors the cluster-wide profile end-to-end, on the wire.

**Scenario 3 — ML-KEM regression check:** confirmed above in both scenarios (`ML-KEM: SUPPORTED`, `X25519MLKEM768`) — this change does not affect the already-working PQC readiness.


abushkin-redhat:
Hey Guzman,

The PR looks good overall. Here are a couple of findings that surfaced during review...

1. **Restart loop on invalid Custom `minTLSVersion` (`cmd/manager/operator.go`)**: `fetchTLSConfig` validates `minTLSVersion` and falls back to defaults on failure, seeding the watcher with those defaults — but the poll path (`SecurityProfileWatcher.Reconcile` → `GetTLSProfileSpec`) doesn't validate `minTLSVersion`, so it reads the raw Custom spec, always sees it differ from the seeded default, and triggers `cancel()`/restart every poll tick, forever. This defeats the intended graceful fallback. A potential solution could be to seed the watcher with what `Reconcile` will actually compute (or re-validating before restarting).

2. **Inverted negative TLS assertions (`tests/framework/utils.go`, `AssertMetricsEndpointRejectsTLSVersion` / `AssertResultServerRejectsTLSVersion`)**: they only fail when `err == nil && !contains("SSL") && !contains("alert")`, but a successful capped handshake makes curl print SSL connection using TLSv1.2, so the check passes and logs "correctly rejected" even when the server actually accepted the connection — the "floor was raised" guarantee isn't really verified. They also treat any `oc run` error (image pull, pod `AlreadyExists`) as a pass. Should invert to fail on a successful handshake and distinguish infra errors from genuine TLS rejection.

...and some more minor nits which can be addressed, but are less important: 

1. Reject helpers reuse fixed pod names (`tls-reject-test` / `rs-tls-reject-test`), which can collide with a still-terminating pod on retry — use unique names per attempt.
2. `extractTLSProfileForTest` only honors `StrictAllComponents`, whereas the operator's `ShouldHonorClusterTLSProfile` also honors unknown/future adherence values — mirror the operator's predicate.
3. `IsOCPVersionAtLeast` reads `Status.History[0]`, which can be a still-applying entry mid-upgrade — prefer the newest `Completed` entry.
4. Consider adding a unit test for the metrics TLS path (as [PR #1059](https://github.com/openshift/file-integrity-operator/pull/1059) has).
5. Clarify whether result-server TLS is in scope — the `AssertResultServer*` helpers are currently unused.

guzalv:
> Hey Guzman,
> 
> The PR looks good overall. Here are a couple of findings that surfaced during review...
> 
> 1. **Restart loop on invalid Custom `minTLSVersion` (`cmd/manager/operator.go`)**: `fetchTLSConfig` validates `minTLSVersion` and falls back to defaults on failure, seeding the watcher with those defaults — but the poll path (`SecurityProfileWatcher.Reconcile` → `GetTLSProfileSpec`) doesn't validate `minTLSVersion`, so it reads the raw Custom spec, always sees it differ from the seeded default, and triggers `cancel()`/restart every poll tick, forever. This defeats the intended graceful fallback. A potential solution could be to seed the watcher with what `Reconcile` will actually compute (or re-validating before restarting).
> 2. **Inverted negative TLS assertions (`tests/framework/utils.go`, `AssertMetricsEndpointRejectsTLSVersion` / `AssertResultServerRejectsTLSVersion`)**: they only fail when `err == nil && !contains("SSL") && !contains("alert")`, but a successful capped handshake makes curl print SSL connection using TLSv1.2, so the check passes and logs "correctly rejected" even when the server actually accepted the connection — the "floor was raised" guarantee isn't really verified. They also treat any `oc run` error (image pull, pod `AlreadyExists`) as a pass. Should invert to fail on a successful handshake and distinguish infra errors from genuine TLS rejection.
> 
> ...and some more minor nits which can be addressed, but are less important:
> 
> 1. Reject helpers reuse fixed pod names (`tls-reject-test` / `rs-tls-reject-test`), which can collide with a still-terminating pod on retry — use unique names per attempt.
> 2. `extractTLSProfileForTest` only honors `StrictAllComponents`, whereas the operator's `ShouldHonorClusterTLSProfile` also honors unknown/future adherence values — mirror the operator's predicate.
> 3. `IsOCPVersionAtLeast` reads `Status.History[0]`, which can be a still-applying entry mid-upgrade — prefer the newest `Completed` entry.
> 4. Consider adding a unit test for the metrics TLS path (as [PR #1059](https://github.com/openshift/file-integrity-operator/pull/1059) has).
> 5. Clarify whether result-server TLS is in scope — the `AssertResultServer*` helpers are currently unused.

Thanks! Addressed in https://github.com/openshift/file-integrity-operator/pull/1065/changes/22b303694413c8d6efd5f607753feea5a239020d, would you like to take another look?

guzalv:
> retest

Looking into the konflux failures

guzalv:
> > retest
> 
> Looking into the konflux failures

Should be fixed in https://github.com/openshift/file-integrity-operator/pull/1065/commits/c58b89df13f405f8bc418c4823cf3ecef15be1f7

vickeybrown (approved):

yuumasato (approved):
/lgtm

PR #1085, opened 2026-09-23
Trim excessive comments added by #1065

guzalv:
## Problem

#1065 added a large number of comments that explain *what* the following code does (restating the code) rather than *why* it exists. That makes the diff much harder to review and adds ongoing maintenance weight to code that will now need those comments kept in sync.

## Change

Trim comments across every file #1065 touched (excluding vendored code) down to only the ones that capture non-obvious rationale (why a value, retry count, ordering, or design choice was made). Removed comments that just restated the following line/function signature (e.g. `// GetOperatorPods returns the operator pods in the operator namespace.`). No functional changes.

Net: 389 lines removed, 137 added.

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./cmd/... ./pkg/controller/metrics/... ./tests/framework/...` — all pass


Vincent056 (approved):
/lgtm

